PR #11, opened 2014-02-25
UDP socket support for Windows Runtime.

MarcelRaad:
This implementation comes with the same caveats as the TCP implementation,
with the exception that explicit binding is supported.

The following socket options are supported:
- IP_TTL
- IPV6_UNICAST_HOPS
- SO_RCVBUF (only Windows 8.1+)
- IP_DONTFRAGMENT (only Windows 8.1+)

It has been tested (as Boost.Asio based on Boost 1.55) with UDPv4
on Visual C++ 2012 (Windows 8.0) and Visual C++ 2013 (Windows 8.1).


mediabuff:
hi
In your start_receive_op implementation is there any reason why you are not following the pattern of TCP socket  - in terms of using the asio async manager.

in TCP we have
async_manager_.async(
        impl.socket_->InputStream->ReadAsync(
          bufs.buffers()[0], bufs.buffers()[0]->Capacity,
          Windows::Storage::Streams::InputStreamOptions::Partial), op);

in UDP I don't see this "queuing for work". This leads to io_service.run to return prematurely.

MarcelRaad:
@mediabuff Sorry, as we have switched to Windows 8.1+ and VS2013.3+, where WinSock can be used instead of the WinRT APIs, I can't test this anymore. I always used this with an io_service that was running for the whole program lifetime.

MarcelRaad:
Closing this as it was for pre-UWP WinRT and I need to get my master branch to mirror upstream master again.

PR #29, opened 2014-09-22
IPv4 addresses in CIDR (Classless Inter Domain Routing) notation 

olk:
class cidr_address_v4 abstracts IPv4 address in CIDR notation providing infos like:
cidr :                          192.168.1.21/28
network (subnet-id) : 192.168.1.16
netmask :                   255.255.255.240
broadcast:                 192.168.1.31
address range:          192.168.1.17 - 192.168.1.30


chriskohlhoff:
Implemented in 956979b4458abd89a7e28d9685f8c178d9ca19fd.

PR #32, opened 2014-10-15
IPv6 addresses in CIDR (Classless Inter Domain Routing) notation

olk:
class network_v6 abstracts IPv6 address in CIDR notation providing infos like:
cidr : 2001:0db8::/119
network : 2001:0db8::
prefix-address : ffff:ffff:ffff:ffff:ffff:ffff:ffff:fe00
address range: 2001:0db8:: - 2001:0db8::01ff


chriskohlhoff:
Implemented in 956979b4458abd89a7e28d9685f8c178d9ca19fd.

PR #40, opened 2014-12-03
Map ConnectEx errors to portable error codes

VemundH:
See https://svn.boost.org/trac/boost/ticket/10744


chriskohlhoff:
Fixed in 33e64f705af47d3393e2239a047c75ab37d1e48f.

PR #42, opened 2014-12-15
Fix error clearing for use_tmp_dh functions

bajamircea:
The SSL error is cleared a bit too late in the functions related to use_tmp_dh.

The change addresses for example the case where in use_tmp_dh_file, the call to BIO_new_file fails.


chriskohlhoff:
Fixed by bdeaf9a277a2460dfe692d67819c158cbf6cd6d1.

PR #48, opened 2015-02-13
move from async_result for coroutines

cmello:
Hi,

These move operations would be handy for me. But I am new to asio, so I really don't know if it may break other code.

Thank you for the attention!

Best regards
Cesar


chriskohlhoff:
Fixed by d35db94c127698310ea9a2040f8ff9f4732a6fb0 and 8d08c7e08e5d53a57f7799dda5ca3be55c1cf987.

PR #58, opened 2015-03-27
Disable __thread keyword for android-clang

crystax:
Clang on Android don't support __thread properly since it require some support from runtime, and Android don't provide it. Look at https://tracker.crystax.net/issues/912 for details. This fix disable usage of __thread in case it's built by clang for Android.

Please include this fix into upcoming Boost 1.58 release (unless there are some objections against this). I'm asking this because this fix is simple and will touch only Android - but it will fix numerous Boost libraries fails, currently listed for Android on http://www.boost.org/development/tests/master/developer/summary.html.


crystax:
OK, as I see, this fix is already [applied to master branch](https://github.com/chriskohlhoff/asio/commit/9c8d15c6b2040f8ad826dd7fb017fb7e21571dff), so closing this PR.

PR #65, opened 2015-05-27
Fix: wrong condition for end check

matbech:
At the point of the proposed change the iter points to the current value (host). We need to check whether the next value is the end. If we don't do that the iterator is incremented and may reach the end. And later the end iterator is dereferenced:
 handler_(static_cast<const asio::error_code&>(ec),
            static_cast<const typename Protocol::endpoint&>(*iter));
This results in an access violation.

### DIFF
--- a/asio/include/asio/impl/connect.hpp
+++ b/asio/include/asio/impl/connect.hpp
@@ -307,7 +307,7 @@ namespace detail
 
           default:
 
-          if (iter == end)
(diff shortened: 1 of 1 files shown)


matbech:
This has been fixed in https://github.com/chriskohlhoff/asio/commit/5b89400ac303ddf8a12b1c541655721a3fc8f505

PR #90, opened 2015-12-25
SSLv3 check

morfin60:
Added check to avoid trying to use deprecated SSLv3 in context.impl when OpenSSL does not have it.


globin:
:+1: And I guess it should normally default to using `SSLv23` to be future compatible.

chriskohlhoff:
Was fixed by 44f3ade708f98a355a322be16985db39b5da5314.

PR #92, opened 2016-01-02
Fix build when only asio/spawn.hpp is included

PSIAlt:
When asio/spawn.hpp is the only one header included, build fails because bind_executor.hpp lacks some includes.


chriskohlhoff:
Fixed by d636aafe8a24e300e9bea891ddb297434d7d98ff.

PR #95, opened 2016-01-21
Fix use of strerror_r()

### LINKED ISSUE #94: build fails with musl libc (on Alpine Linux) due to falling back to non-standard strerror_r()
Build fails with this error:
````
In file included from ./../../../include/asio/error_code.hpp:185:0,
                 from ./../../../include/asio/io_service.hpp:25,
                 from ./../../../include/asio/basic_io_object.hpp:19,
                 from ./../../../include/asio/basic_socket.hpp:20,
                 from ./../../../include/asio/basic_datagram_socket.hpp:20,
                 from ./../../../include/asio.hpp:19,
                 from allocation/server.cpp:19:
./../../../include/asio/impl/error_code.ipp: In member function 'virtual std::__cxx11::string asio::detail::system_category::message(int) const':
./../../../include/asio/impl/error_code.ipp:110:22: error: could not convert 'strerror_r(value, ((char*)(& buf)), sizeof (buf))' from 'int' to 'std::__cxx11::string {aka std::__cxx11::basic_string<char>}'
     return strerror_r(value, buf, sizeof(buf));
                      ^
````

The problem is that code seems to consider the non-standard GNU implementation of strerror_r as the standard and make all known platforms (which follows the [posix standard](http://pubs.opengroup.org/onlinepubs/9699919799/functions/strerror.html)) as exceptions.

### PR DESCRIPTION
**Pull request by ncopa**
Don't try keep track of all platforms which follows posix standard.
Instead, make the platform which don't follow standard to an exception
and fall back to standard implementation for everything else.

This fixes building with musl libc.

Fixes #94

### DIFF
--- a/asio/include/asio/impl/error_code.ipp
+++ b/asio/include/asio/impl/error_code.ipp
@@ -98,17 +98,14 @@ public:
 #if defined(__sun) || defined(__QNX__) || defined(__SYMBIAN32__)
     using namespace std;
     return strerror(value);
-#elif defined(__MACH__) && defined(__APPLE__) \
-  || defined(__NetBSD__) || defined(__FreeBSD__) || defined(__OpenBSD__) \
-  || defined(_AIX) || defined(__hpux) || defined(__osf__) \
-  || defined(__ANDROID__)
+#elif defined(__GLIBC__) && defined(_GNU_SOURCE)
+    char buf[256] = "";
+    return strerror_r(value, buf, sizeof(buf));
+#else
     char buf[256] = "";
     using namespace std;
     strerror_r(value, buf, sizeof(buf));
     return buf;
-#else
-    char buf[256] = "";
-    return strerror_r(value, buf, sizeof(buf));
 #endif
 #endif // defined(ASIO_WINDOWS_DESKTOP) || defined(__CYGWIN__)
   }

### CODE REVIEW

File: asio/include/asio/impl/error_code.ipp [hunk 1]
```diff
@@ -98,17 +98,14 @@ public:
 #if defined(__sun) || defined(__QNX__) || defined(__SYMBIAN32__)
     using namespace std;
     return strerror(value);
-#elif defined(__MACH__) && defined(__APPLE__) \
-  || defined(__NetBSD__) || defined(__FreeBSD__) || defined(__OpenBSD__) \
-  || defined(_AIX) || defined(__hpux) || defined(__osf__) \
-  || defined(__ANDROID__)
+#elif defined(__GLIBC__) && defined(_GNU_SOURCE)
```
Soullivaneuh: Musl library is not gnu but also have strerror_r implemented.

http://git.musl-libc.org/cgit/musl/tree/include/string.h#n64

I don't think it will work with `defined(_GNU_SOURCE)` for this case.

ncopa:
Musl libc has `strerror_r` becuase it is psecified in [posix](http://pubs.opengroup.org/onlinepubs/9699919799/functions/strerror.html). But strerror_r() is really nasty. There are 2 different versions of strerror_r: the posix standard and the GNU. The function have same name but does different things and returns different things.

from http://man7.org/linux/man-pages/man3/strerror.3.html:
````
       int strerror_r(int errnum, char *buf, size_t buflen);
                   /* XSI-compliant */

       char *strerror_r(int errnum, char *buf, size_t buflen);
                   /* GNU-specific */
...

   strerror_r()
       The strerror_r() function is similar to strerror(), but is thread
       safe.  This function is available in two versions: an XSI-compliant
       version specified in POSIX.1-2001 (available since glibc 2.3.4, but
       not POSIX-compliant until glibc 2.13), and a GNU-specific version
       (available since glibc 2.0).  The XSI-compliant version is provided
       with the feature test macros settings shown in the SYNOPSIS;
       otherwise the GNU-specific version is provided.  If no feature test
       macros are explicitly defined, then (since glibc 2.4) _POSIX_SOURCE
       is defined by default with the value 200112L, so that the XSI-
       compliant version of strerror_r() is provided by default.

       The XSI-compliant strerror_r() is preferred for portable
       applications.  It returns the error string in the user-supplied
       buffer buf of length buflen.

       The GNU-specific strerror_r() returns a pointer to a string
       containing the error message.  This may be either a pointer to a
       string that the function stores in buf, or a pointer to some
       (immutable) static string (in which case buf is unused).  If the
       function stores a string in buf, then at most buflen bytes are stored
       (the string may be truncated if buflen is too small and errnum is
       unknown).  The string always includes a terminating null byte ('\0').
````

My patch will check for the non-standard GNU variant for the single known platform that uses it (GNU) and assume that all other platforms, including musl libc, uses the standard variant.

Soullivaneuh:
Oh ok, I didn't see the difference. That's quite vicious.

Thanks for the tip, I'll try to apply it on mongo project. :+1:

chriskohlhoff:
Fixed in 443bc17d13eb5e37de780ea6e23157493cf7b3b9. Please note I took a different approach which eliminated those particular #ifdefs.

Issue #101, opened 2016-03-07
Is it really necessary to have a global mutex protecting calls to SSL_accept?


dkadashev:
`SSL_accept()` call inside of `ssl::detail::engine::do_accept()` is protected by a global mutex. Not surprisingly this is a huge bottleneck.

OpenSSL doc says that "Provided an application sets up the thread callback function" OpenSSL is thread-safe (https://www.openssl.org/docs/faq.html#PROG1). I think the mutex was originally added here: https://sourceforge.net/p/asio/patches/1/ - could it be the case that back then either thread locking callback was not configured [properly] or there was an error in testing? Does the OpenSSL doc lie?

The mutex contention heavily limits achievable request rates.

acmorrow:
Only somewhat related, but related nonetheless is https://github.com/chriskohlhoff/asio/issues/73.

dkadashev:
We did several stress tests on an 8-core box with patched ASIO that doesn't protect `SSL_accept()` with mutex. No crashes / errors / memory leaks observed after ~65M requests at ~1350 requests per second (CPU was maxed out, mostly by SSL code).

Looks like the mutex is not necessary as OpenSSL docs suggest.

dkadashev:
Checked BoringSSL docs just in case, and there is no indication that `SSL_accept()` is not thread safe there as well. I think the mutex can be removed unconditionally.

byllyfish:
According to this bug report from 2010, the first call to SSL_accept() from two different threads could lead to a race condition. This bug is supposed to be fixed.

https://rt.openssl.org/Ticket/Display.html?id=2208

From ssl.h:
> An SSL object represents a single TLS or DTLS connection. Although the shared SSL_CTX is thread-safe, an SSL is not thread-safe and may only be used on one thread at a time.

dkadashev:
So that bug report is about first `SSL_accept()` call only. But the mutex is locked always. It's too high price to pay if you ask me. Also it's a long fixed bug. I wouldn't say it's an asio responsibility to work around OpenSSL bugs.

As for SSL object not being thread safe, it's not relevant here. The mutex in question is global, not per-SSL object. And it's not asio responsibility to prevent users from doing stupid things like using single SSL object concurrently. Note that only `SSL_accept()` is protected by the mutex, but most of the `engine` methods operate on SSL object. If making `engine` objects thread safe was part of the design then (almost) every method operating on `ssl_` would have to be protected by a mutex (not a global one).

[...]

PR #102, opened 2016-03-10
Remove global mutex protecting SSL_accept() calls

dkadashev:
There is no indication in OpenSSL / BoringSSL docs that SSL_accept() is
not thread safe if "an application sets up the thread callback
function". ASIO does set up those functions. So SSL_accept() should be
thread safe.

The mutex was suggested by https://sourceforge.net/p/asio/patches/1/.
Probably there was a mistake in user testing (e.g. thread callback
function was not configured). Regardless, the global mutex locking leads
to heavy lock contention and thus very low performance. In our tests we
used to have ~200-250 requests per second, and CPU was far from being
maxed out.

We did some stress tests with the mutex removed from the asio code.
After ~65M requests at ~1350 req/sec there was no indication of any
problems (no crashes / errors / memory leaks). The test box CPU was
maxed out. Looks like the documentation does not lie and the mutex is
not necessary.

### DIFF
--- a/asio/include/asio/ssl/detail/engine.hpp
+++ b/asio/include/asio/ssl/detail/engine.hpp
@@ -115,10 +115,6 @@ private:
   ASIO_DECL static int verify_callback_function(
       int preverified, X509_STORE_CTX* ctx);
 
-  // The SSL_accept function may not be thread safe. This mutex is used to
-  // protect all calls to the SSL_accept function.
-  ASIO_DECL static asio::detail::static_mutex& accept_mutex();
-
   // Perform one operation. Returns >= 0 on success or error, want_read if the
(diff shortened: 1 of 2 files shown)


dkadashev:
This is supposed to be a fix for #101

PSIAlt:
Confirmed RPS issue in multithreaded app with this lock (and even slower with perfect-forwarding ciphers).
Took the header into our project in order to remove this mutex, and it works well with heavy load.

chriskohlhoff:
Fixed in 37d0d474f2562d9ec50025ecacd3613e9e406685. Conservatively, I have retained the locking for openssl < 1.0.

xiaopeifeng:
thx the fix

PR #104, opened 2016-03-11
Don't rely on deprecated implicit OpenSSL includes

MarcelRaad:
rsa.h and dh.h are needed in context.ipp for RSA_free and DH_free.
They are only included implicitly if OPENSSL_NO_DEPRECATED is not defined.


MarcelRaad:
https://github.com/chriskohlhoff/asio/commit/69e44a4cc6eb5ba21ede409779a7b777c0eb3869

PR #109, opened 2016-03-14
Fix configure.ac to support mingw64

rpavlik:
This enhances the configure script to recognize MinGW64, as present in MSYS2, as a valid Windows platform, and thus to link with the right Windows libraries, and to not try building syslog daemons and such on it.


chriskohlhoff:
Fixed on master in 13aff8506778957252b831370dff5357f336e994. Fixed on asio-1-10-branch in b5d411192df97eb9683765f2a5970496f2e80f16.

Issue #108, opened 2016-03-14
Can't make check with ssl support in MinGW64/msys2


rpavlik:
It appears to be because of transitive includes: one of the OpenSSL header files ends up including `winsock.h` (not `winsock2.h` - they are incompatible), so it all falls apart in `socket_types.h`

I extracted the include tree with this command, mildly modified from the build command make was running:

``` bash
( cd src/tests/ &&  \
  /mingw64/bin/g++ -DPACKAGE_NAME=\"asio\" -DPACKAGE_TARNAME=\"asio\" -DPACKAGE_VERSION=\"1.11.0\" -DPACKAGE_STRING=\"asio\ 1.11.0\" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -DPACKAGE=\"asio\" -DVERSION=\"1.11.0\" -D_REENTRANT=1 -DSTDC_HEADERS=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_MEMORY_H=1 -DHAVE_STRINGS_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_UNISTD_H=1 -I.    -I./../../include -g -O2 -mthreads -ftemplate-depth-256 \
  unit/ssl/context_base.cpp -M -H ) \
  > sslincludes.txt 2>&1
```

Here's the subset of that `-M -H` output of trying to build the test that led me to the flaw: (empty lines mean I left out un-related lines - lines that weren't direct parents)

```
.. ./../../include/asio/ssl/detail/openssl_types.hpp

... C:/msys2/mingw64/include/openssl/ssl.h

.... C:/msys2/mingw64/include/openssl/dtls1.h
..... C:/msys2/mingw64/include/openssl/pqueue.h
...... C:/msys2/mingw64/x86_64-w64-mingw32/include/string.h
....... C:/msys2/mingw64/x86_64-w64-mingw32/include/sec_api/string_s.h
........ C:/msys2/mingw64/x86_64-w64-mingw32/include/string.h
..... C:/msys2/mingw64/x86_64-w64-mingw32/include/winsock.h
```

so it's the `openssl/dtls1.h` file at fault here - and they're just including it for `struct timeval` - oh the humanity!

[...]

PR #110, opened 2016-03-14
On Windows, if using SSL, include socket_types before OpenSSL headers.

rpavlik:
This is the non-pretty solution to #108: it does work to complete a build successfully, and it's not really that gross of a hack, it's just platform-specificity leaking where you'd probably rather not have it, but OpenSSL forced the hand.


chriskohlhoff:
Fixed. See #108 for details.

PR #120, opened 2016-06-13
Changed include paths to be relative and no longer require adding an …

damu:
…includepath to the compiler options. Not sure if I changed every path correctly but the async_tcp_echo_server example works for me now.
Having relative paths makes bundling a library much easier and build scrips as well.


chriskohlhoff:
The intention is that the asio `include` directory be added to your project settings, e.g. by using the compiler's `-I` command line option. All `#include` directives are relative to that point.

PR #123, opened 2016-06-24
add operator+ to mutable_buffers_1/const_buffers_1

zhuoqiang:
While there's `operator+` for `mutable_buffer`, however there's no corresponding `operator+` for `mutable_buffers_1` which makes `read` into buffer at specific offset position harder.

For example, the following code yields compile error: 
```
boost/boost/asio/detail/consuming_buffers.hpp:261:21: error: no type named 'const_iterator' in
      'boost::asio::mutable_buffer'
  typename Buffers::const_iterator begin_remainder_;
  ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
```
```c++
  std::array<char, 64> buf;
  boost::asio::io_service ios;
  boost::asio::ip::tcp::socket socket(ios);
  boost::asio::async_read(
      socket, 
      boost::asio::buffer(buf)+1, [](auto ec, auto size) {});
```
it is because operator+ only return `mutable_buffer` while `async_read` requires `mutable_buffers_1`

In order to get a buffer with corret offset position, user has to write:
```c++
auto b = boost::asio::mutable_buffer(buf.data()+1, buf.size()-1);
```
or
```c++
auto b = asio::buffer(asio::buffer(buf) + 1);
```

If asio could provide `operator+` for `mutable_buffers_1`, the above code could be simplified to
```
auto b = boost::asio::buffer(buf) + 1;
```

### DIFF
--- a/asio/include/asio/buffer.hpp
+++ b/asio/include/asio/buffer.hpp
@@ -520,6 +520,63 @@ inline const_buffer operator+(std::size_t start,
   return b + start;
 }
 
+/// Create a new modifiable buffer that is offset from the start of another.
+/**
+ * @relates mutable_buffers_1
+ */
+inline mutable_buffers_1 operator+(const mutable_buffers_1& b,
+                                std::size_t start) ASIO_NOEXCEPT
+{
+  if (start > b.size())
+    return mutable_buffers_1(NULL, 0);
+  char* new_data = static_cast<char*>(b.data()) + start;
+  std::size_t new_size = b.size() - start;
+  return mutable_buffers_1(new_data, new_size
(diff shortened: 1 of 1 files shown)


chriskohlhoff:
The buffer sequence requirements have changed, as of commit 16120e1bd7eb5db62b8acc4c1d51f313fa805234, such that asio::const_buffer and asio::mutable_buffer satisfy the requirements directly. For example. this means that the following code now works:

    boost::asio::async_read(
        socket, 
        boost::asio::buffer(buf)+1, [](auto ec, auto size) {});

The asio::const_buffers_1 and asio::mutable_buffers_1 classes have been deprecated.

PR #127, opened 2016-07-22
Recognize loopback IPv4-mapped IPv6 addresses as such

eric135:
Per https://tools.ietf.org/html/rfc4291#section-2.5.5.2, IPv4 addresses can be mapped to IPv6 addresses (`127.0.0.1` turns into `[::ffff:127.0.0.1]`). Since `127.0.0.0/8` addresses are loopback, I would assume that their IPv6 mappings would be loopback as well.

This issue has been previously reported on the Boost bug tracker (https://svn.boost.org/trac/boost/ticket/9084). I believe I've coded a solution to it.

### DIFF
--- a/asio/include/asio/ip/impl/address_v6.ipp
+++ b/asio/include/asio/ip/impl/address_v6.ipp
@@ -147,7 +147,8 @@ bool address_v6::is_loopback() const
       && (addr_.s6_addr[8] == 0) && (addr_.s6_addr[9] == 0)
       && (addr_.s6_addr[10] == 0) && (addr_.s6_addr[11] == 0)
       && (addr_.s6_addr[12] == 0) && (addr_.s6_addr[13] == 0)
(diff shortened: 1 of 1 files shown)


yoursunny:
Is this the correct repository? There's a https://github.com/boostorg/asio as well.

eric135:
From what I understand, the code is actually developed in this repo and then "boostified" to the Boost.Asio repo. Please correct me if I'm wrong.

yoursunny:
http://www.boost.org/development/pull_requests.php is linking to https://github.com/boostorg/asio

eric135:
I've created a pull request on the Boost.Asio repo as well (https://github.com/boostorg/asio/pull/42). It appears I was incorrect, as that repo has more recent commits. The two pull requests should be more or less the same and the maintainer can consider whichever pull request is in the correct repo.

PR #133, opened 2016-09-12
Fix forward declaration with defaulted args

matbech:
The compiler does not like this syntax:
``
  typename Time = chrono::steady_clock,
    typename TimeTraits = wait_traits<Time>
``
Compiler error:
experimental/__net_ts/ts/netfwd.hpp(130): error C2065: 'Time': undeclared identifier

This patch works around the problem.


matbech:
This has been fixed with https://github.com/chriskohlhoff/asio/commit/8737ab0f2fa259a4f0cb08a46257fee24914a1b3

PR #136, opened 2016-09-24
Fix for 'vector iterator + offset out of range'

bsergeev:
`std::advance(end, next_elem_);` causes 'vector iterator + offset out of range' in Debug build.


rpavlik:
This change is the same as #152, fwiw.

chriskohlhoff:
Fixed on master in bc6e494d2d58b491f0ccec7113060eb10192b781.

PR #138, opened 2016-10-06
Acceptex

rpavlik:
Turns out that directly linking to AcceptEx is as bad of an idea in disparate environments as ConnectEx (failed in at least one of my test build environments). This code handles the AcceptEx case the same way as ConnectEx was already handled, in part by factoring out and re-using some of the connectex machinery.


chriskohlhoff:
I tried to merge this but it didn't work. I suspect the GetAcceptExSockaddrs function also needs to be looked up, for this to work portably across all versions of Windows.

PR #140, opened 2016-10-10
fix #139: fix apple's broken getaddrinfo

### LINKED ISSUE #139: Could not connect to numeric port in Apple's NAT64 environment
Apple suggests to using `getaddrinfo()` to synthesize IPv6 address in Apple's NAT64 environment. 

However, Apple's `getaddrinfo()` is broken if numeric port is used as the service name. It always return port 0 in this case. So asio always failed to get the right endpoint for connection.

zhuoqiang:
Apple suggests to using `getaddrinfo()` to synthesize IPv6 address in Apple's NAT64 environment. 

However, Apple's `getaddrinfo()` is broken if numeric port is used as the service name. It always return port 0 in this case. So asio always failed to get the right endpoint for connection.

The fix is to correct the numeric port after `getaddrinfo()` call.

See: https://forums.developer.apple.com/thread/47312

Fix #139

### DIFF
--- a/asio/include/asio/detail/impl/socket_ops.ipp
+++ b/asio/include/asio/detail/impl/socket_ops.ipp
@@ -3327,6 +3327,38 @@ asio::error_code getaddrinfo(const char* host,
   return ec = translate_addrinfo_error(error);
 #else
   int error = ::getaddrinfo(host, service, &hints, result);
+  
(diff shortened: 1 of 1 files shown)


chriskohlhoff:
Fixed. See #139 for details.

PR #142, opened 2016-10-15
Re-enable building against LibreSSL

ennui93:
After 062b19c97bb85f4625b46f93ee19b234948ff235, I had some issues building programs making use of boost against LibreSSL on my system. I found that this change fixed those programs' build.

Disclaimer: I am neither a LibreSSL developer, nor am I familiar with OpenSSL APIs, nor do I even know if this is the 'right' fix.

### DIFF
--- a/asio/include/asio/ssl/impl/context.ipp
+++ b/asio/include/asio/ssl/impl/context.ipp
@@ -215,7 +215,7 @@ context::~context()
 {
   if (handle_)
   {
-#if (OPENSSL_VERSION_NUMBER >= 0x10100000L)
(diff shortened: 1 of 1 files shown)


MarcelRaad:
Sorry for the breakage. LibreSSL was forked from OpenSSL 1.0.1 but pretends to be OpenSSL 2, so this fix looks right to me. curl uses the same workaround (https://github.com/curl/curl/blob/master/lib/vtls/openssl.c).

chriskohlhoff:
I believe both master and asio-1-10-branch now build against LibreSSL.

PR #145, opened 2016-10-17
Problem of hangs caused by the infinite wait or constant readiness to read the new data

sysprg:
Recently I encountered a problem of hangs caused by the infinite wait in numerous unit tests during build of the Galera project from the Codership.

In particular, this problem manifests itself in the following configurations:

- debian-wheezy-x64 (https://wiki.debian.org/DebianWheezy);
- ubuntu-wily-x64 (http://old-releases.ubuntu.com/releases/15.10/);
- debian-jessie-x64 with AddressSanitizer (https://www.debian.org/releases/jessie/ + https://en.wikipedia.org/wiki/AddressSanitizer);
- centos6-32 (https://wiki.centos.org/Manuals/ReleaseNotes/CentOS6.8);

In particular, the program may hang in the epoll_wait() function, or vice versa, this function can be completed instantly (on other systems).

I think this error occurs in the ASIO C++ library because the Linux kernel requires the application to read an 8-byte timeout expiration counter (using the "timerfd" descriptor) every time when the EPOLLIN event occurs. Otherwise, the timer stops to notifying the application about new events, and all activity associated with this timer will be frozen on some systems, or vice versa, we get an infinite loop of notifications because of the constant readiness to read the new data.

But current version of the ASIO library does not read the data using the "timerfd" descriptor after returning from the epoll_wait().

To fix this problem, we should read the 8-byte expiration counter from the "timerfd" descriptor every time, when the EPOLLIN event occurs.

### DIFF
--- a/asio/include/asio/detail/impl/epoll_reactor.ipp
+++ b/asio/include/asio/detail/impl/epoll_reactor.ipp
@@ -503,6 +503,24 @@ void epoll_reactor::run(long usec, op_queue<operation>& ops)
 #if defined(ASIO_HAS_TIMERFD)
     else if (ptr == &timer_fd_)
     {
+      // We should read an 8-byte timeout expiration counter (using
+      // the "timerfd" descriptor) every time when the EPOLLIN event
+      // occurs. Otherwise, the timer stops to notifying the application
+      // about new events, and all activity associated with this timer
+      // will be frozen on some systems, or vice versa, we get an infinite
+      // loop of notifications because of the constant readiness to read
+      // the new data:
+
+      if (events[i].events & EPOLLIN)
+      {
+        // If timeout was expired, we should read the expiration counter:
+        uint64_t count;
+        if (read(timer_fd_, &count, sizeof(count)))
+        {
+          // Just to ignore return value of the read() function
+          // without compiler warning...
+        }
+      }
       check_timers = true;
     }
 #endif // defined(ASIO_HAS_TIMERFD)


asio/include/asio/detail/impl/epoll_reactor.ipp
```diff
@@ -503,6 +503,24 @@ void epoll_reactor::run(long usec, op_queue<operation>& ops)
...
+      // about new events, and all activity associated with this timer
+      // will be frozen on some systems, or vice versa, we get an infinite
+      // loop of notifications because of the constant readiness to read
+      // the new data:
+
+      if (events[i].events & EPOLLIN)
+      {
+        // If timeout was expired, we should read the expiration counter:
+        uint64_t count;
+        if (read(timer_fd_, &count, sizeof(count)))
```
2underscores-vic: What if `read()` fails with `EINTR`?

sysprg: After returning from the epoll_wait() with EPOLLIN notification the read() system call is never blocked, so it never returns EINTR in the real world. And, since timerfd_create() returns pseudo-device handle (instead of the real device), then read() never return less than 8 bytes. But formally, for reassurance, we can write some code like this:
```
ssize_t len = sizeof(count);
char * buf = reinterpret_cast<char *>(&count);
do {
  ssize_t ret = read(timer_fd_, buf, len);
  if (ret < 0)
  {
    If (EINTR != errno) break;
    continue;
  }
  buf += ret;
  len -= ret;
} while (len);
```

sysprg: After returning from the epoll_wait() with EPOLLIN notification the read() system call is never blocked, so it never returns EINTR in the real world. And, since timerfd_create() returns pseudo-device handle (instead of the real device), then read() never return less than 8 bytes. But formally, for reassurance, we can write some code like this:
```
ssize_t len = sizeof(count);
char * buf = reinterpret_cast<char *>(&count);
do {
  ssize_t ret = read(timer_fd_, buf, len);
  if (ret < 0)
  {
    If (EINTR != errno) break;
    continue;
  }
  buf += ret;
  len -= ret;
} while (len);
```

2underscores-vic: What about making the code "more optimistic" for performance? :-)

```
for(;;) {
    ssize_t ret = read(timer_fd_, buf, len);
    if(ret == len) break;
    ...
```

chriskohlhoff:
This needs a program to reproduce the issue.

N.B. According to the timerfd_settime man page:

> If  the timer has already expired one or more times **_since its settings were last modified using timerfd_settime()_**, or since the last successful read(2), then
> the buffer given to read(2) returns an unsigned 8-byte integer (uint64_t) containing the number of expirations that have occurred.  (The returned value is in
> host byte order—that is, the native byte order for integers on the host machine.)

The bottom of `epoll_reactor::run()` always calls `timerfd_settime` if `check_timers` is set to true. Thus the man page indicates that the `read` would be redundant.

PR #148, opened 2016-10-31
Fix compilation with OpenSSL 1.1 API

MarcelRaad:
With OPENSSL_API_COMPAT=0x10100000L, SSL_library_init, SSL_load_error_strings, and OpenSSL_add_all_algorithms are removed.
With OPENSSL_API_COMPAT=0x10000000L, these are function-style macros mapping to OPENSSL_init_ssl, which is called automatically anyway.

References:
https://www.openssl.org/docs/man1.1.0/ssl/OPENSSL_init_ssl.html
https://www.openssl.org/docs/man1.1.0/crypto/OPENSSL_init_crypto.html

### DIFF
--- a/asio/include/asio/ssl/detail/impl/openssl_init.ipp
+++ b/asio/include/asio/ssl/detail/impl/openssl_init.ipp
@@ -35,11 +35,11 @@ class openssl_init_base::do_init
 public:
   do_init()
   {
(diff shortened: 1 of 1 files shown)


chriskohlhoff:
Fix applied in 4fa65a5f7710cd075cf53533fdaab91f5e5506ab.

PR #149, opened 2016-11-01
Use std::atomic_thread_fence on Mac OSX >= 10.12 and iOS >= 10

esteve:
This pull request removes a deprecation warning that occurs when building Asio for Mac OSX 10.12 or iOS 10, by using C++11's `std::atomic_thread_fence` instead of `OSMemoryBarrier`

Addresses https://github.com/chriskohlhoff/asio/issues/144

Issue #144, opened 2016-10-17
Use of deprecated OSMemoryBarrier on latest MacOS


jeremyong:
asio/asio/include/asio/detail/macos_fenced_block.hpp:51:5: warning: 'OSMemoryBarrier' is deprecated: first deprecated in macOS 10.12 - Use std::atomic_thread_fence() from <atomic> instead [-Wdeprecated-declarations]
    OSMemoryBarrier();
    ^
/usr/include/libkern/OSAtomicDeprecated.h:749:9: note: 'OSMemoryBarrier' has been explicitly marked deprecated here
void    OSMemoryBarrier( void );

jeremyong:
Actually seems like a number of things are deprecated and no longer working in latest MacOS (xcode 8.1 compiler)

twn346:
There was an update to XCode recently, and using the header-only version that does not rely on Boost the only warning I got was this one about OSMemoryBarrier. 

Compiled with: -std=c++14 -Wall -Wextra -pedantic -g -stdlib=libc++

chriskohlhoff:
I have applied a fix that addresses the warning when compiling for C++11 or later. No idea what we're supposed to do for C++03.

Fixed on master in ffc9dd443fc2ca1aece0bbda14f3c521ceeb7924. Fixed on asio-1-10-branch in 99c0f9239e61e215b21b226a3432e1440710df01.

PR #151, opened 2016-11-02
Use std::atomic_thread_fence if <atomic> exists

esteve:
This pull request is an alternate approach to #149 , instead of specializing `macos_fenced_block` for C++11, I added a new fenced block implementation for toolchains that have `<atomic>`

This is a bit less conservative than #149 since it just assumes that any toolchain with `<atomic>` will provide `std::atomic_thread_fence`, though I don't know if there's one where the latter does not exist but still implements the rest of the C++11 atomic API.

Addresses #144

### DIFF
--- a/asio/include/asio/detail/atomic_fenced_block.hpp
+++ b/asio/include/asio/detail/atomic_fenced_block.hpp
@@ -1,0 +1,62 @@
+//
+// detail/atomic_fenced_block.hpp
+// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+//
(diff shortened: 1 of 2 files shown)


esteve:
Deleting the accompanying branch since this PR was merged into master in https://github.com/chriskohlhoff/asio/commit/ffc9dd443fc2ca1aece0bbda14f3c521ceeb7924 and into the asio-1-10-branch in https://github.com/chriskohlhoff/asio/commit/99c0f9239e61e215b21b226a3432e1440710df01. Thanks @chriskohlhoff

PR #152, opened 2016-11-05
Fix consuming buffer.

mmicko:
Fix consuming buffer. Fixes Visual Studio crash and transfer of larger buffers works fine now.

Example for C++11 version of http server was transfering just up to 64k files and with this change all working fine.


chronoxor:
Also found this issue. Cannot wait the fix will be merged!

mmicko:
Fixed in https://github.com/chriskohlhoff/asio/commit/bc6e494d2d58b491f0ccec7113060eb10192b781

PR #158, opened 2016-11-17
Reset parser state machine between requests

didenko:
The state machine in HTTP Server C++11 example was not reset between parsing of requests. Which makes second and later requests fail if the previous request left the parser in an unfortunate state.


didenko:
Never mind, the example does not support neither keep-alive connections not HTTP pipelining - so the reset is irrelevant.

PR #169, opened 2016-12-09
Use allocator_traits to rebind allocators.

### LINKED ISSUE #168: create_promise uses C++03 allocator rebinding
```
    p_ = std::allocate_shared<std::promise<T>>(
        typename Allocator::template rebind<char>::other(a),
        std::allocator_arg,
        typename Allocator::template rebind<char>::other(a));
```

This assumes that the allocator has a nested `rebind` template, which is true for `std::allocator` but not for C++11 allocators in general. It should use `allocator_traits<Allocator>::rebind_alloc<char>`.

(This is partly my fault, for breaking the use of `allocator<void>::construct` so that rebinding is needed in the first place, which is now fixed in GCC svn, see https://gcc.gnu.org/PR78052 - sorry!)

jwakely:
Also add value_type typedef to hook_allocator.

Fixes #168

### DIFF
--- a/asio/include/asio/detail/recycling_allocator.hpp
+++ b/asio/include/asio/detail/recycling_allocator.hpp
@@ -29,6 +29,8 @@ template <typename T>
 class recycling_allocator
 {
 public:
+  typedef T value_type;
+
   template <typename U>
   struct rebind
   {
@@ -62,6 +64,8 @@ template <>
 class recycling_allocator<void>
 {
 public:
+  typedef void value_type;
+
   template <typename U>
   struct rebind
   {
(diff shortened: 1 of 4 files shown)


asio/include/asio/detail/recycling_allocator.hpp
```diff
@@ -29,6 +29,8 @@ template <typename T>
 class recycling_allocator
 {
 public:
+  typedef T value_type;
+
   template <typename U>
   struct rebind
```
jwakely: This `rebind` template (and the one in `recycling_allocator<void>`) could even be removed, because `allocator_traits` will do the right thing even without it.

chriskohlhoff:
Fixed, see #168 for details.

Issue #14, opened 2014-04-28
select_reactor::run ignores error_code


Urmeli0815:
In select_reactor.ipp, line 212, the outparam "ec" (error_code) of the socket_ops::select call is ignored.

If I understand the code correctly this can lead to a busy loop when one of the underlying sockets is damaged/invalid. We experienced high CPU usage in our application by one specific thread, especially when the process was idle for a long time or woke from hibernation. The callstack always lead to the select_reactor-Loop.

Some more information: our application makes a client connection to a partner process via async_connect. This internally opens three socket connections for the "self pipe trick" (to support cancellation of asynchronous connections). When I close one of the two "trick" sockets from outside (e. g. via TCPView) the thread wreaks havoc using up one core.

As far as I understand the "trick" sockets are only needed as long as the real connection is not established. If this is the case it would probably be better to clean them up when the connection is established.

MarcelRaad:
I often have the same problem, but with error_code 0. The select call wakes up immediately, 0 bytes are read in interrupter_.reset(), and the thread burns one CPU core.

MarcelRaad:
Fixed by #171 .

PR #171, opened 2016-12-13
Close old sockets and open new ones to watch when recv fails

timmaraju:
When the ec is ignored, the same socket fds are passed to select over and over again, which returns them as to be read causing an endless loop. [https://github.com/chriskohlhoff/asio/issues/14](url)


chriskohlhoff:

PR #173, opened 2016-12-18
Tv titles can also be considered windows apps

baszalmstra:


chriskohlhoff:

PR #176, opened 2017-01-11
InitializeCriticalSectionEx has undocumented max spincount of 0xFFFFFF

bradhugh:
On Win 10 Universal Windows (Store) apps InitializeCriticalSectionEx was failing with 0x00000057 ERROR_INVALID_PARAMETER.  Found that there is apparently a max value for this function on Win10 of 0xFFFFFF.  Also added error handling that wasn't there.


chriskohlhoff:
Fixed in ced6124d2c1f3f56cb9c131ba8634bf1b367ae54.

Issue #175, opened 2017-01-09
basic_streambuf not movable


BMBurstein:
I am not sure why `basic_streambuf` is marked as `noncopyable`, but can it at least be made movable?

MarcelRaad:
I just stumbled across the same problem today when I wondered why `std::unique_ptr<asio::streambuf>` is used in so many places.

2underscores-vic:
Similar issue - #124

vinniefalco:
Check out this substitute for `basic_streambuf`
https://github.com/vinniefalco/Beast/blob/master/include/beast/core/basic_streambuf.hpp#L21

PR #179, opened 2017-01-17
Make noncopyable still be movable

BMBurstein:
Fix for #175


chriskohlhoff:
Sorry, this change is not valid.

BMBurstein:
Why is it invalid? And is there another solution for the problem I was trying to fix?

chriskohlhoff:
It's invalid because it globally changes things that aren't intended to be movable, movable.

The issue in #175 requires a change specifically for basic_streambuf.

PR #192, opened 2017-03-12
Add config options to disable examples and tests

joerg-krause:
I am preparing to add asio to Buildroot as a package. If selected by other packages as a dependency, it is not necessary to have the examples and tests build. Therefore, add config options `--disable-examples` and `--disable-tests` to speed up the build process and having only the headers installed in a staging directory.


joerg-krause:
No feedback. Closing.

PR #208, opened 2017-05-26
Replace shared_ptr with unique_ptr

nunojpg:
This pull request is obviously incomplete as it only updates one example, but I would like to get comments if a shared_ptr had actually any business here, of if unique_ptr is clearly the right choice.


vinniefalco:
if it is half-duplex then `unique_ptr` can work but otherwise it should be `shared_ptr`.

PR #210, opened 2017-06-11
is_continuation misplaced in reactive socket service

arun11299:
The argument `is_continuation` passed to `start_op` should be the third-last instead of second-last.


chriskohlhoff:
Fixed in 48e7ef864f66f6e8403be2fb4c1f63d21a5902e6.

Issue #215, opened 2017-06-22
master regression from 1.10.6 with coroutine spawn and async_resolve


cedral:
This code works in 1.10.6 but in master it decrements outstanding_work_ to zero and stops the io service (despite the existence of a work object for the io service) when the coroutine completes. It only does this if I use async_resolve. I am using visual studio 2017 on windows. But the same thing happens on MacOS using clang from Xcode 8.1

```
#include <thread>
#include <memory>
#include <future>
#include <iostream>
#include <asio.hpp>
#include <asio/spawn.hpp>

using namespace asio;
using namespace asio::ip;

std::future<void> Connect(tcp::socket &s, const char *host, unsigned short port)
{
	std::packaged_task<void(yield_context)> task{
		[host,port,&s](yield_context yield)
		{
			tcp::resolver::query _query(host, std::to_string(port), tcp::resolver::query::canonical_name);
			tcp::resolver _resolver{ s.get_io_service() };
			auto _iterator = _resolver.async_resolve(_query, yield);
		}
	};
	auto future = task.get_future();
	asio::spawn(s.get_io_service(), std::move(task));
	return future;
}

int main()
{
	io_service service;
	tcp::socket socket(service);
	std::unique_ptr<io_service::work> work_{ new io_service::work{ service } };
	std::thread t{ [&] {service.run(); } };
	Connect(socket, "serengeti.com", 80).get();
	std::cout << "this is as far as we get" << std::endl;
	service.post([]() { std::cout << "mostly don't get here" << std::endl; });
	work_.reset();
	t.join();
	std::cin.get();
	return 0;
}
```

chriskohlhoff:
Fixed in b8a34a1dc0285259408365695119dd29747fcd34, thanks.

PR #217, opened 2017-06-29
resolver_query resulting in one 2 many derefs of outstanding_work_

cedral:
This fixed my problem in issue #215 
It derrefs just the right amount with this patch. tried a loop of 50 resolves and io_context.run() finishes when I reset my work guard at the end. Sorry about the spacing getting messed up. That wasn't visible in visual studio. This actually only moving one line.


chriskohlhoff:
Fixed in b8a34a1dc0285259408365695119dd29747fcd34.

PR #218, opened 2017-07-03
This pull request allows people to select boost::future over std::future

cedral:
I needed continuation with std::future doesn't have.
simple
#define ASIO_DISABLE_STD_FUTURE 
and it ill fail over to boost::future unless ASIO_DISABLE_BOOST_FUTURE is also defined.


cedral:
The exception handling wasn't working right. boost::promise doesn't handle std::exception_ptr well.

cedral:
I also added support for passing in an error_code * to use_future_t so that you can retrieve an error without throwing an exception on a call to get() or co_await

cedral:
This works as well as I can make it. The integration tests are now failing because of the 
#include <tuple>
and use of std::tuple on non c++11 compilers. I didn't do that so I can't fix it. If you don't want to support those compilers perhaps you should remove them from your tests? There are only 3 in there.

PR #221, opened 2017-07-12
Avoid exception throwing from `make_network_v{4,6}(str, ec)`

gohyda:


chriskohlhoff:
Fixed in ddef4994f42f2eb3aa1e4b8f376d3e42facfaa4b.

gohyda:
This PR is not completely fix the issue: with invalid prefix-length the problem still exists. 
#265 fix this.

PR #231, opened 2017-07-29
Only enable string_view with /std:c++17 on MSVC

### LINKED ISSUE #230: string_view and VS 2017
Just including these two : 

```C++
#include "asio/detail/config.hpp"
#include "asio/detail/string_view.hpp"
```

fails with 
```
#error directive: string_view is only available with C++17
```
(i.e. one has to build with /std:c++17 or maybe /std:c++latest, which is not always possible: for instance currently this breaks boost due to the use of deprecated `std::unary_function` stuff in it).
 
Interestingly, including either one or the other works, as well as changing the order of the includes.

```C++
// ok: 
#include "asio/detail/string_view.hpp"
#include "asio/detail/config.hpp"
```

a solution could be to guard the `#include <string_view>` with `defined(_HAS_CXX17)` since it seems that's the macro MS uses in their string_view header.

jcelerier:
Fixes #230

### DIFF
--- a/asio/include/asio/detail/config.hpp
+++ b/asio/include/asio/detail/config.hpp
@@ -790,7 +790,7 @@
 #   endif // ((__GNUC__ == 4) && (__GNUC_MINOR__ >= 7)) || (__GNUC__ > 4)
 #  endif // defined(__GNUC__)
 #  if defined(ASIO_MSVC)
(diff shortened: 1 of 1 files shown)


chriskohlhoff:
Fixed in d6144828beb342548eaf7b31132217247cb03e27.

PR #242, opened 2017-08-30
Add file descriptors to epoll in first call to start_op

### LINKED ISSUE #240: Accepting sockets asynchronously apparently adversely affects future synchronous IO performance on some platforms
I'm filing this issue perhaps prematurely, because I do not yet have a reduced reproduction. I plan to work on one, but I wanted to put this up first in case this was a known issue or is somehow expected behavior.

The setup: The system was originally written such that there was a TCP/IPv4 accept socket, and new connections were accepted by a dedicated thread (the "listen" thread), that would do blocking calls to `acceptor::accept`. Once a connection was accepted, the socket for the new connection was handed to a new dedicated thread for that connection, which proceeded to do blocking network IO on the socket, and the listen thread would return to its blocking call to `accept`.

There was, however, a problem with this model, since there is no portable way to unblock a blocking call to `accept`. Calling `close` from another thread is famously unsafe, and calling `shutdown` has unspecified semantics.

To address this, the accept path was re-written to have a dedicated thread doing non-blocking accepts via an `io_service` object and calls to `acceptor::async_accept`. After connections were accepted, they were handed off to the same backend, with a new thread for the connection doing blocking network IO.

This fixed the issue with unblocking from accept, since we could now `cancel` the call to `async_accept` safely. However, our CI system flagged the commit as causing a significant performance regression in some of our benchmarks; in some cases, as high as 30%.

This was somewhat shocking, because the benchmark in question simply opens a fixed number of connections, waits for them all to be accepted, and then drives load over the already accepted connections. It does not create new connections during the test, and measures throughput only during the portion of the test during which all connections are active.

The upshot is that it *seems* that the mechanism of accepting a socket affects the future performance of the accepted socket. A socket accepted via the non-blocking `acceptor::async_accept` performs much worse than a socket accepted via `acceptor::accept`.

We are using ASIO 230c0d2ae035c5ce1292233fcab03cea0d341264. The system where this is being tested is a CentOS 6 Linux instance.

Note also that I've had no luck reproducing these results on my local workstation, running Ubuntu 16.04, but the change in behavior is very observable on our performance benchmark system. Switching the accept mechanism back to a blocking call to `acceptor::accept` definitively restores performance on the benchmarking hardware.

I realize this is not the most actionable bug report, and I'm hoping to provide more detail in the future, but I would be interested to hear any thoughts. This issue is potentially blocking the rollout of a new ASIO based ingress networking layer in MongoDB.

jbreams:
This fixes #240.

Since sockets accepted by async_accept may only be used in synchronous I/O, they shouldn't be added to epoll until the first time an async op is done. Otherwise threads running the io_context will wake up from epoll_wait every time any file descriptor becomes readable, even though there are no async ops waiting for it.

### DIFF
--- a/asio/include/asio/detail/epoll_reactor.hpp
+++ b/asio/include/asio/detail/epoll_reactor.hpp
@@ -64,6 +64,7 @@ public:
     op_queue<reactor_op> op_queue_[max_ops];
     bool try_speculative_[max_ops];
     bool shutdown_;
+    bool added_to_epoll_;
 
     ASIO_DECL descriptor_state(bool locking);
     void set_ready_events(uint32_t events) { task_result_ = events; }
--- a/asio/include/asio/detail/impl/epoll_reactor.ipp
+++ b/asio/include/asio/detail/impl/epoll_reactor.ipp
@@ -162,27 +162,9 @@ int epoll_reactor::register_descriptor(socket_type descriptor,
     descriptor_data->reactor_ = this;
     descriptor_data->descriptor_ = descriptor;
     descriptor_data->shutdown_ = false;
+    descriptor_data->added_to_epoll_ = false;
     for (int i = 0; i < max_ops; ++i)
       descriptor_data->try_speculative_[i] = true;
-  }
-
-  epoll_event ev = { 0, { 0 } };
-  ev.events = EPOLLIN | EPOLLERR | EPOLLHUP | EPOLLPRI | EPOLLET;
-  descriptor_data->registered_events_ = ev.events;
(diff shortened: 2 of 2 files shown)


acmorrow:
A gentle ping on this? We are still interested in continuing the conversation, and still feel that our proposed resolution is superior to the guidance to use a separate reactor. If we are mistaken, we would like to understand why. For reference, please see my comments in #240.

acmorrow:
@chriskohlhoff - Any thoughts on this one? We still are interested in it, please see the conversation in #240.

PR #249, opened 2017-09-09
Construct the associated allocator only when needed

vinniefalco:
This fixes a bug where the implementation attempts to construct the associated allocator for a handler whose storage was already deallocated.


vinniefalco:
Looks like this was added as https://github.com/chriskohlhoff/asio/commit/2ccb648f44c50c6f1002fabfd3c305a15785242c

It would have been nice to use a real C++ declaration instead of putting code in the macro, since the macro technique interferes with debugging in Visual Studio.

PR #264, opened 2017-11-01
Fix includes in some cpp11 and cpp14 examples

podsvirov:
This small changes fix building of the some standalone examples with MSVC.

Includes:
* ctime - for std::time_t and std::ctime();
* numeric - for std::iota();
* cctype - for std::toupper().

I tested this with MSVC2017.


chriskohlhoff:
Applied in d31062f3fca6d3e11ccf1c485f55cfdcffd2d66e.

podsvirov:
Thanks :-)

PR #266, opened 2017-11-06
Let spawn support lvalue Handler and tie Function life time

yayj:
Please check the full testing code at [Gist](https://gist.github.com/yayj/8aa00e55bc85d030c7205b54112c14e7#file-pr266-cpp)

- no compilation error when passing a lvalue handler to spawn
```
auto handler = [] {};
asio::spawn(handler, [](auto) {});  // spawn_data cannot be created since template parameter Handler is not a universal reference
```
- no runtime abortion when passing a temporary function
```
asio::io_service io;
{
  auto pf = make_unique<void(*)(asio::yield_context)>([](auto) {});
  asio::spawn(io, *pf);
}
io.run();     // runtime error since Function reference has been invalid.
```

### DIFF
--- a/asio/include/asio/impl/spawn.hpp
+++ b/asio/include/asio/impl/spawn.hpp
@@ -351,6 +351,10 @@ namespace detail {
   template <typename Handler, typename Function>
   struct spawn_data : private noncopyable
   {
+    typedef typename decay<Handler>::type handler_type;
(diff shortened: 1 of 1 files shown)


chriskohlhoff:
Likely addressed by commit 939f25a9e, released in asio 1.23.0.

Issue #268, opened 2017-11-25
AddrSan - heap-use-after-free due to double-free of Object*, possibly in epoll_reactor


Olipro:
Somewhere *after* commit hash 524288, I've had addrsan complaining on a particular test about a heap-use-after-free. It appears to be caused by a double-free since if I edit `object_pool.hpp` so that the `free()` function checks if `o == free_list_` the problem goes away. Without the change, we end up with the object in the free list having `next->` pointing to itself.

Here's the AddrSan output:

`    ==40103==ERROR: AddressSanitizer: heap-use-after-free on address 0x60f000007ee8 at pc 0x0000006484ad bp 0x7ffe75219ba0 sp 0x7ffe75219b98
    READ of size 8 at 0x60f000007ee8 thread T0
        #0 0x6484ac in asio::detail::object_pool<asio::detail::epoll_reactor::descriptor_state>::destroy_list(asio::detail::epoll_reactor::descriptor_state*) /home/oliver/Projects/asioproj/build/debug/asio/asio/include/asio/detail/object_pool.hpp:154:14
        #1 0x6461bd in asio::detail::object_pool<asio::detail::epoll_reactor::descriptor_state>::~object_pool() /home/oliver/Projects/asioproj/build/debug/asio/asio/include/asio/detail/object_pool.hpp:78:5
        #2 0x646465 in asio::detail::epoll_reactor::~epoll_reactor() /home/oliver/Projects/asioproj/build/debug/asio/asio/include/asio/detail/impl/epoll_reactor.ipp:70:1
        #3 0x6464d8 in asio::detail::epoll_reactor::~epoll_reactor() /home/oliver/Projects/asioproj/build/debug/asio/asio/include/asio/detail/impl/epoll_reactor.ipp:65:1
        #4 0x56ea28 in asio::detail::service_registry::destroy(asio::execution_context::service*) /home/oliver/Projects/asioproj/build/debug/asio/asio/include/asio/detail/impl/service_registry.ipp:109:3
        #5 0x56e95f in asio::detail::service_registry::destroy_services() /home/oliver/Projects/asioproj/build/debug/asio/asio/include/asio/detail/impl/service_registry.ipp:53:5
        #6 0x56e73c in asio::execution_context::destroy() /home/oliver/Projects/asioproj/build/debug/asio/asio/include/asio/impl/execution_context.ipp:45:22
[...]

PR #269, opened 2017-11-25
Don't do free() if Object has already been freed

Olipro:
Somewhere within ASIO, a double-free is occurring. Since I unfortunately can't pin it down further, this check ensures we don't free the object twice.

See Issue #268 for further details.


vinniefalco:
Seems related to https://github.com/boostorg/asio/commit/aa0bf0a8b34815cbf4ecac9e36c2a809385bd458

vinniefalco:
Is this still an issue?

PR #277, opened 2018-02-07
Fix issue with string_view under Clang 5.0

chronoxor:


chriskohlhoff:
Addressed by 3c7bbe42ea6768b885ae0408edcd80d6e76d0173.

PR #280, opened 2018-02-15
add const to basic_socket_acceptor::get_option()

inetknght:
...since it appears to be const-safe and would fix an issue I encountered during use


chriskohlhoff:
Fixed by 7b2d186cedc01864edb16d1827b7dd729e36086f.

PR #285, opened 2018-02-26
Fix deprecated use of async_result

inetic:
For easier transition.


chriskohlhoff:
Applied in 21e86ec26d0f06d571afe6f9d509bc546b67d86b.

PR #291, opened 2018-03-14
Correct prepare_memfn_helper test

regular384:
Fix prepare_memfn_helper test to use intended member function
'prepare' rather than 'data' copied by mistake from previous check.


chriskohlhoff:
Applied in 68180a19ff4f1e3b0270524aecb936de353a2a9b.

PR #294, opened 2018-03-19
Make add_certificate_authority process multiple certificates in a bundle

bsergeev:
Currently, `ssl::context::add_certificate_authority()` processed only a single certificate from a bundle passed as its 1st parameter. My change allowes it to process all certificates.


chriskohlhoff:
Applied in 1156fad4f0aeda946d397024fe0900afa8d95e84.

PR #296, opened 2018-04-02
add conan package setup for standalone version of asio

ClausKlein:
The coanan package is created with cmake and tested with ctest


ClausKlein:
I added a .clang-tidy file too. It can be used with cmake while compileing or as post compile check.
see too:
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines
https://www.kdab.com/clang-tidy-part-1-modernize-source-code-using-c11c14/
https://clang.llvm.org/extra/clang-tidy/index.html

PR #299, opened 2018-04-23
config: Fix condition for NOEXCEPT to accept ASIO_STANDALONE

scosu:
The project including this header may be using boost but asio as a
standalone. In this case BOOST_NOEXCEPT_OR_NOTHROW may not be resolved
properly and fail.

This patch checks the boost version only if ASIO_STANDALONE is not set.

Signed-off-by: Markus Pargmann <scosu@quobyte.com>


chriskohlhoff:
Fixed by 83e24fa04f75371b0c033197e959eedebe816b4d.

PR #301, opened 2018-04-28
Use std::move in basic_resolver_results::begin() to avoid copying

petrhosek:
This is reported as `-Wreturn-std-move` warning by recent versions of Clang.


chriskohlhoff:
Applied in 912d0d67f9de0b130fc274003ef883854b4e2f54.

PR #317, opened 2018-06-28
Fix bug in deregister_handler

Bklyn:
Code as written does not correctly de-register handlers.


chriskohlhoff:
Applied in 6554073823c5b2afa380a5d5960ce2c693eb5cd6.

PR #319, opened 2018-07-02
config.hpp: added support for Intel Compiler std::move (fixes #257)

### LINKED ISSUE #257: Intel C++ Compiler feature detection
Move semantics support (and possibly other C++11 features) in Intel C++ is currently not detected in asio/asio/include/asio/detail/config.hpp

I tested compilation with Intel C++ 18 (__INTEL_COMPILER=1800)

From my understanding Intel C++ tries to track what MSVC and GCC support on Windows/Linux

I am far from being an expert in predefined macros by various compilers but looking at config.hpp I think that this part needs to be changed (for Windows platform):

```
// Microsoft Visual C++ detection.
#if !defined(ASIO_MSVC)
# if defined(ASIO_HAS_BOOST_CONFIG) && defined(BOOST_MSVC)
#  define ASIO_MSVC BOOST_MSVC
# elif defined(_MSC_VER) && (defined(__INTELLISENSE__) \
      || (!defined(__MWERKS__) && !defined(__EDG_VERSION__)))
#  define ASIO_MSVC _MSC_VER
# endif // defined(ASIO_HAS_BOOST_CONFIG) && defined(BOOST_MSVC)
#endif // defined(ASIO_MSVC)
```

`ASIO_HAS_BOOST_CONFIG) && defined(BOOST_MSVC)
`
Fails since BOOST_MSVC is undefined when Intel C++ is detected

```
_MSC_VER) && (defined(__INTELLISENSE__) \
      || (!defined(__MWERKS__) && !defined(__EDG_VERSION__)))
```
Fails because  `__INTELLISENSE__`  is defined only when IntelliSense parses the code

> __INTELLISENSE__ Defined as 1 during an IntelliSense compiler pass in the Visual Studio IDE. Otherwise, undefined. You can use this macro to guard code the IntelliSense compiler does not understand, or use it to toggle between the build and IntelliSense compiler.

And because  `__EDG_VERSION__` is defined as 413 (this macro is not defined when compiling with MSVC)

In my opinion the check for `__INTELLISENSE__` does not make any sense, since its purpose is for hiding code from IntelliSense.

emmenlau:
Added support to detect the std::move capabilities of the Intel Compiler. I am not certain about the version number, the estimate may be a bit pessimistic, so earlier compiler versions may already support std::move. I could not reliably detect this in the Intel documentation.
Fixes issue #257

### DIFF
--- a/asio/include/asio/detail/config.hpp
+++ b/asio/include/asio/detail/config.hpp
@@ -111,6 +111,14 @@
 #    define ASIO_HAS_MOVE 1
 #   endif // (_MSC_VER >= 1700)
 #  endif // defined(ASIO_MSVC)
+#  if defined(__INTEL_CXX11_MODE__)
+#    if defined(__INTEL_COMPILER) && (__INTEL_COMPILER >= 1500)
+#      define BOOST_ASIO_HAS_MOVE 1
+#    endif // defined(__INTEL_COMPILER) && (__INTEL_COMPILER >= 1500)
+#    if defined(__ICL) && (__ICL >= 1500)
+#      define BOOST_ASIO_HAS_MOVE 1
+#    endif // defined(__ICL) && (__ICL >= 1500)
+#  endif // defined(__INTEL_CXX11_MODE__)
 # endif // !defined(ASIO_DISABLE_MOVE)
 #endif // !defined(ASIO_HAS_MOVE)
 


emmenlau:
Is this an acceptable change or can I do something to support it? Currently asio does not correctly work with the Intel compiler.

chriskohlhoff:
Applied in 547f06a1e4662fdbe6728ad18699163798c41dc2.

emmenlau:
Thanks!

PR #322, opened 2018-07-19
thread_pool utilize ASIO_CONCURRENCY_HINT_1 when (num_threads == 1)

stalkyr:


chriskohlhoff:
Fixed on master, but using `1` rather than `ASIO_CONCURRENCY_HINT_1`.

PR #323, opened 2018-07-21
Replace OSMemoryBarrier() atomic fence

powderluv:
Since macOS 10.12 OSMemoryBarrier() is deprecated so replace it with
std::atomic_thread_fence()

### DIFF
--- a/asio/include/asio/detail/macos_fenced_block.hpp
+++ b/asio/include/asio/detail/macos_fenced_block.hpp
@@ -19,6 +19,7 @@
 
 #if defined(__MACH__) && defined(__APPLE__)
 
+#include <atomic>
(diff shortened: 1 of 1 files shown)


asio/include/asio/detail/macos_fenced_block.hpp
```diff
@@ -19,6 +19,7 @@
 
 #if defined(__MACH__) && defined(__APPLE__)
 
+#include <atomic>
 #include <libkern/OSAtomic.h>
```
dimitry-ishenko: FWIW this header is no longer needed (since you removed all uses of `OSMemoryBarrier()`)

chriskohlhoff:
Addressed in asio version 1.13.0.

PR #341, opened 2018-11-02
Do not double define has_string_view with clang-cl

dutow:
Clang-cl identifies both as clang and as MSVC.
This resulted in defining ASIO_HAS_STD_STRING_VIEW twice.


chriskohlhoff:
Applied in 0f9b727e9f22b93c12b13e38e4998e2c33a2bfd8.

PR #342, opened 2018-11-02
Check for and enable local Unix sockets on Windows.

amensel:
Rough draft of enabling Unix domain socket support on Windows, when possible.

It's not full feature set, yet, but it's more than enough to be useful: https://blogs.msdn.microsoft.com/commandline/2017/12/19/af_unix-comes-to-windows/


chriskohlhoff:

PR #347, opened 2018-11-16
Add support for configuring symbol visibility using macro

rahulj09:
As of now all the symbols are exported by asio , since in push_options.hpp we always set symbol visibility to default i.e.https://github.com/chriskohlhoff/asio/blob/master/asio/include/asio/detail/push_options.hpp#L27

We have a requirement where we don't want to export symbols from our dylib but since asio always exports symbols we are facing issues.

### DIFF
--- a/asio/include/asio/detail/pop_options.hpp
+++ b/asio/include/asio/detail/pop_options.hpp
@@ -23,7 +23,7 @@
 
 // Intel C++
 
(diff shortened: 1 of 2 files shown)


chriskohlhoff:
A similar change has been applied in 1c9a4ac3386b543d1608ed929d12f8ae0f63d8d8.

PR #351, opened 2018-11-28
Let asio build on Haiku OS

jackburton79:
Hi, I added some changes to be able to build ASIO on Haiku OS.
Currently all tests pass, except 3: multicast, v6only and socket_base.

Will investigate these problems, but I think the first two fail because of missing support in the network stack,


chriskohlhoff:
Applied in 0616f498bfa5796ae28016cccc1bff78fce04c26 and 38d3fc09ab69f5bf870f6616f3fdec67674becd3.

PR #353, opened 2018-12-17
Correctly enabling multicast on QNX

MiguelCompany:
As QNX has a BSD-styled reuse address/port behavior, this small patch adds SO_REUSEPORT for QNX systems.


chriskohlhoff:
Applied in a10d3595eb33bb6cb75199533c5178a2b8676332.

PR #360, opened 2019-01-28
return correct size when a truncation occures

Zyrin:
Applied changes discussed a while ago (in May 2018) via email.

The problem as about boost::asio::ip::udp::socket::receive_from() returning size == 0 and error == false in the truncation case.


chriskohlhoff:
Applied in 555ed6c993113be78e14a51fc26a6ee8ad996e62.

PR #361, opened 2019-02-02
Added constructor for local::basic_endpoint from string_view

abbyssoul:
This PR add a new constructor to asio::local::basic_endpoint<> that takes string_view. 
It appears there is no need to use std::string or NULL-terminated string. This PR also brings creation options of local::basic_endpoint on par with ip::address make_address.


chriskohlhoff:
Applied in 171a02ed0de69356a48aa587115f6c3e47f48563.

PR #366, opened 2019-02-12
Make the executor_work_guard move constructor noexcept

RedDwarf69:
If I got it right https://www.boost.org/doc/libs/1_69_0/doc/html/boost_asio/reference/Executor1.html says the executors are guaranteed to be move-noexcept and so the executor_work_guard is free to also be?


chriskohlhoff:
Applied in 3b0daafa0b713c6621746969dc9ffdf65b1158f6.

PR #399, opened 2019-06-04
Remove unnecessary copy when looping iterators

wyattoday:
The logic is identical, avoids a copy of an iterator, and uses less code.


chriskohlhoff:
This change is not correct. The body of each loop may result in removal of the element, so we must take a copy and increment the iterator first.

PR #402, opened 2019-07-02
Test for is_buffer_sequence: false positives

karzhenkov:
Simple test cases indicate that the `is_buffer_sequence` template is implemented incorrectly:
- `struct A1` with single member is recognized as a _MutableBufferSequence_ when compile with the `decltype` available;
- `struct B1` with inappropriate typedefs and no member functions is recognized as a _MutableBufferSequence_ when the `decltype` is unavailable or disabled.

### DIFF
--- a/asio/include/asio/detail/is_buffer_sequence.hpp
+++ b/asio/include/asio/detail/is_buffer_sequence.hpp
@@ -214,8 +214,8 @@ char mutable_buffers_type_typedef_helper(
 template <typename T, typename Buffer>
 struct is_buffer_sequence_class
   : integral_constant<bool,
-      sizeof(buffer_sequence_begin_helper<T>(0)) != 1 &&
-      sizeof(buffer_sequence_end_helper<T>(0)) != 1 &&
+      sizeof(buffer_sequence_begin_helper<T>(0, 0)) != 1 &&
+      sizeof(buffer_sequence_end_helper<T>(0, 0)) != 1 &&
       sizeof(buffer_sequence_element_type_helper<T, Buffer>(0, 0)) == 1>
 {
 };
--- a/asio/include/asio/detail/is_buffer_sequence.hpp
+++ b/asio/include/asio/detail/is_buffer_sequence.hpp
@@ -53,18 +53,21 @@ struct buffer_sequence_memfns_check
 {
 };
 
-template <typename>
-char (&buffer_sequence_begin_helper(...))[2];
-
 #if defined(ASIO_HAS_DECLTYPE)
 
+template <typename>
+char buffer_sequence_begin_helper(...);
+
 template <typename T>
-char buffer_sequence_begin_helper(T* t,
+char (&buffer_sequence_begin_helper(T* t,
     typename enable_if<!is_same<
       decltype(asio::buffer_sequence_begin(*t)),
-        void>::value>::type*);
+        void>::value>::type*))[2];
 
 #else // defined(ASIO_HAS_DECLTYPE)
+
+template <typename>
+char (&buffer_sequence_begin_helper(...))[2];
 
 template <typename T>
 char buffer_sequence_begin_helper(T* t,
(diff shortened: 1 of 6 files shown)


asio/include/asio/detail/is_buffer_sequence.hpp
```diff
@@ -74,19 +77,22 @@ char buffer_sequence_begin_helper(T* t,
...
+template <typename>
+char buffer_sequence_end_helper(...);
+
 template <typename T>
-char buffer_sequence_end_helper(T* t,
+char (&buffer_sequence_end_helper(T* t,
     typename enable_if<!is_same<
       decltype(asio::buffer_sequence_end(*t)),
-        void>::value>::type*);
+        void>::value>::type*))[2];
```
raJeev-M: Why this change?

asio/include/asio/detail/is_buffer_sequence.hpp
```diff
@@ -214,8 +220,8 @@ char mutable_buffers_type_typedef_helper(
 template <typename T, typename Buffer>
 struct is_buffer_sequence_class
   : integral_constant<bool,
-      sizeof(buffer_sequence_begin_helper<T>(0)) != 1 &&
-      sizeof(buffer_sequence_end_helper<T>(0)) != 1 &&
+      sizeof(buffer_sequence_begin_helper<T>(0, 0)) != 1 &&
+      sizeof(buffer_sequence_end_helper<T>(0, 0)) != 1 &&
```
raJeev-M: Now the fallback becomes `char[2]` not `char, so should it not be `!=2`

asio/include/asio/detail/is_buffer_sequence.hpp
```diff
@@ -53,19 +53,22 @@ struct buffer_sequence_memfns_check
...
+template <typename>
+char buffer_sequence_begin_helper(...);
+
 template <typename T>
-char buffer_sequence_begin_helper(T* t,
+char (&buffer_sequence_begin_helper(T* t,
     typename enable_if<!is_same<
       decltype(asio::buffer_sequence_begin(*t)),
-        void>::value>::type*);
+        void>::value>::type*))[2];
```
raJeev-M: why this change?

asio/include/asio/detail/is_buffer_sequence.hpp
karzhenkov: When called with single argument this is revolved to the fallback, since no the default argument value in checker function declaration is specified. The modified fallback for `ASIO_HAS_DECLTYPE` case is `char`, so `!= 1` is correct.

When `ASIO_HAS_DECLTYPE` is not defined, fallback is still `char[2]`, but in this case checker function must be rejected for acceptable type; `!= 1` is correct again.

asio/include/asio/detail/is_buffer_sequence.hpp
karzhenkov: Fallback is now `char` when `ASIO_HAS_DECLTYPE` is defined (see above).
For checker function we need `char[2]`

raJeev-M:
> struct A1 with single member is recognized as a MutableBufferSequence when compile with the decltype available;

I am not sure why do you say it, Have looked at your test-code, here is my observation and correct me if am wrong

so you have the below stuff

```
struct A1
{
  mutable_buffer* begin();

  // no "value_type" type
  // no "const_iterator" type
  // no "end" member function
};

ASIO_CHECK(!is_mutable_buffer_sequence<A1>::value);
```
I believe ASIO_CHECK returns true.. and it bombs you out because there is no 
`end()`, yet, still the `buffer_sequence_end_helper` is not fallen-back?

Here is my experiment on `asio::buffer_sequence_end_helper`,  looks like, it works fine as expected: https://godbolt.org/z/Ch_jCq . you think I miss anything?

karzhenkov:
When I add `mutable_buffer* end();` to `struct A1` in your sample, it still compile successfully. The second overload of helper (i. e. the checker) will always be rejected independently of declarations in `struct A1`. Condition will always be true.

To change this we can provide default value for its second argument (or add second argument when it is called from `static_assert`).

After such a change we need also invert condition, and write `!= 2`, but this will break the logic for the case when `ASIO_HAS_DECLTYPE` is not defined.

So instead of `!= 2` we have to "swap" return types of the fallback and the checker.

karzhenkov:
When ASIO_HAS_DECLTYPE | for acceptable type the checker must be | fallback return type before change | fallback return type after change
-|-|-|-
is defined | accepted (but was rejected) | `char[2]` | `char`
is not defined | rejected | `char[2]` | `char[2]`

karzhenkov:
>```
> ASIO_CHECK(!is_mutable_buffer_sequence<A1>::value);
> ```
> 
> I believe `is_mutable_buffer_sequence<A1>` returns true.. and it bombs you out because there is no
> `end()`, ...

I would interpret this situation a bit differently.

The test (055c51fc448ef2ce14ef9a357af2309826b374cd) is written with assumption  that `A1` should **not** be considered as _MutableBufferSequence_ because it doesn't have `end()`. The test  [failed](https://ci.appveyor.com/project/chriskohlhoff/asio/builds/25697760/job/92tobk9arn49i605), i. e. the program wrongly accepted `A1` as a valid _MutableBufferSequence_. This indicates an error in the `is_buffer_sequence` implemenation.

Or perhaps the assumption used in the test is wrong?

karzhenkov:
Obsoleted by #403

PR #403, opened 2019-07-03
Fix is_buffer_sequence

karzhenkov:
Almost completely reworked `is_buffer_sequence` implementation is intended to fix issues from #402 (and to satisfy the corresponding test). It is also more concise, since it does not introduce several helper declarations for each individual check. A few helper declarations of general use are placed in the separate header file.

Original logic for all types except `mutable_buffer` and `const_buffer` is guessed as follows:
- When `decltype` is available, the type `X` is considered as _MutableBufferSequence_ if the result of the expression `*begin_buffer_sequence(x)` can be converted to `mutable_buffer` and `end_buffer_sequence(x)` returns non-void. Said functions in general case return the same as corresponding member functions, so the reqiurements above can be applied to expressions `*x.begin()` and `x.end()`.
- When `decltype` is unavailable or disabled, the type `X` is considered as _MutableBufferSequence_ if it has thr following: members named `begin` and `end` (not nessesary the functions); nested types or typedefs `const_iterator` and `value_type`. The latter type also has to be convertible to `mutable_buffer`.

Reqiurements for _ConstBufferSequence_ are similar.

The logic described is implemented incorrectly; test cases and necessary fixes are proposed in #402.
This logic is also not perfect. Some obvious checks in #402 still fail.

The proposed variant implements the following set of requirements for _MutableBufferSequence_ and _ConstBufferSequence_ (with no dependency on the avaliability of `decltype'):

1. Expressions `*begin_buffer_sequence(x)` and `*end_buffer_sequence(x)` have to be convertible to the corresponding buffer type (`mutable_buffer` or `const_buffer`).
1. Expressions `*x.begin()` and `*x.end()` have to be convertible to the corresponding buffer type.
1. `X::value_type` has to be convertible to the corresponding buffer type.

Nested type `const_iterator` (and also `iterator`) will be checked indirectly through declarations of `begin_buffer_sequence` and `end_buffer_sequence`. Ideally, all the necessary requirements besides the first one would be also checked through these functions. In particular, explicit instantiations of `is_buffer_sequence` template for `mutable_buffer` and `const_buffer` would not be needed.


raJeev-M:
Have not looked at the implementation yet, will do so, but before that.

> When decltype is available...
>When decltype is unavailable...

>The proposed variant implements.....(with no dependency on the avaliability of `decltype' )

Are these statements not contradicting themselves? from the first two quote we can clearly see that there is no dependency (currently) on `decltype`.

karzhenkov:
Hi, @raJeev-M 
There is currently a dependency on `decltype`. I tried to describe it in "original logic" part. Perhaps I did it not quite clear, however you can see the difference when run the test from #402 (very first commit 055c51fc448ef2ce14ef9a357af2309826b374cd) with `decltype` enabled/disabled (I used `ASIO_DISABLE_DECLTYPE).

raJeev-M:
> Hi, @raJeev-M
> There is currently a dependency on `decltype`. I tried to describe it in "original logic" part. Perhaps I did it not quite clear, however you can see the difference when run the test from #402 (very first commit [055c51f](https://github.com/chriskohlhoff/asio/commit/055c51fc448ef2ce14ef9a357af2309826b374cd)) with `decltype` enabled/disabled (I used `ASIO_DISABLE_DECLTYPE).

Still, I don't think I will use the word `dependency` , in `is_buffer_sequence` I see it says `ASIO_HAS_DECLTYPE`: which sounds like optional not dependent.

karzhenkov:
I agree, `decltype` is optional; the word "_dependency_" is probably inappropriate here.
The probem is that
- the program behavior differs significantly when compiled with/without this option;
- in both cases the program behavior may be incorrect.

karzhenkov:
`is_dynamic_buffer` templates are improved to more closely meet the documented _DynamicBuffer_ requirements (and also to be a bit more readable).

asio/include/asio/detail/is_buffer_sequence.hpp
```diff
@@ -252,49 +92,76 @@ struct is_buffer_sequence<const_buffer, mutable_buffer>
...
-      sizeof(const_buffers_type_typedef_helper<T>(0)) == 1 &&
-      sizeof(mutable_buffers_type_typedef_helper<T>(0)) == 1>
+struct is_dynamic_buffer_class_v1 : sfinae_check_base
 {
+  static const std::size_t arg = 0;
+
+  template <typename T>
+  static result<sizeof(
+
+    is_same_as<std::size_t>(declval<T>().size()),
```
karzhenkov: Perhaps we should make this check (and other similar checks) more strict:
```suggestion
    is_same_as<std::size_t>(declval<const T>().size()),
```
And also to relax the return type requirement:
```suggestion
    is_convertible_to<std::size_t>(declval<T>().size()),
```

PR #410, opened 2019-07-16
Update type_traits.hpp

jmcruz-uma:
corrected two confusing comments with wrong macro  ASIO_HAS_TYPE_TRAITS to ASIO_HAS_STD_TYPE_TRAITS
**came here looking for an error with asio-standalone 1.30 and gcc9 after including <asio.hpp> in source code: cannot open source file "boost/type_traits/add_const.hpp" (dependency of "asio.hpp")


chriskohlhoff:

PR #433, opened 2019-10-17
NetBSD adapted its kqueue interface to other operating systems'.

0-wiz-0:
Starting with NetBSD 9.99.14, they are the same as in FreeBSD and others.


chriskohlhoff:

PR #437, opened 2019-11-07
Start handler work again in case of restarting accept operation

ewienik:
When connection is aborted during accept operation, before restart_accept_op() handler_work should be restarted also. In other case win_iocp_io_context.outstanding_work_ is not incremented as it should be.


chriskohlhoff:
Addressed in asio version 1.16.0.

PR #438, opened 2019-11-09
rfc2818 validation

arvidn:
instead of implementing certificate hostname pattern matching in rfc2818_verification, rely on X509_check_ip_asc and X509_check_host from openssl.

As far as I can tell, this requires openssl 1.0.2 ( [docs](https://www.openssl.org/docs/man1.0.2/man3/X509_check_host.html) )


chriskohlhoff:
Addressed in asio version 1.16.1, with a new class `host_name_verification` that supersedes `rfc2818_verification`.

PR #440, opened 2019-11-28
Using zero-copy API in order to move data between SSL buffers and the t…

romange:
…ransport layer.

I would like to hear maintainers opinion whether to continue with this.


romange:
@chriskohlhoff  can you please take a look? .

PR #450, opened 2020-01-07
Fix unit test for is_*_buffer_sequence

karzhenkov:
The following causes a compiler error when `ASIO_DISABLE_DECLTYPE` is defined (or when `decltype` is unavailable):

	valid_mutable_a x;
	buffer_sequence_begin(x);

Here `valid_mutable_a` is a type defined in the unit test and approved by `is_const_buffer_sequence` and `is_mutable_buffer_sequence` templates.


chriskohlhoff:

PR #458, opened 2020-03-06
Fix work counting issue in asynchronous resolver endpoint implementation.

jyxjrh:
Fix work counting issue in asynchronous resolver endpoint implementation.


chriskohlhoff:
This was fixed in asio 1.16.1 / boost 1.73.

PR #466, opened 2020-03-20
Fixed bug of async_read_until not completing immediately

mpusz:
Fixed bug of async_read_until not completing immediately when a provided `streambuf` already contains the delimiter.

### DIFF
--- a/asio/include/asio/impl/read_until.hpp
+++ b/asio/include/asio/impl/read_until.hpp
@@ -1118,6 +1118,7 @@ namespace detail
               // Full match. We're done.
               search_position_ = result.first - begin + delim_.length();
               bytes_to_read = 0;
+              start = 0;
             }
 
             // No match yet. Check if buffer is full.


Boilerguy:
Hi, I have a couple of things to comment:
1. There are several places to have the same flow in the file except the place you patched. Should we fix them too?
2. The "start" is the argument passed into. So if we assign a value to it, I think we need another variable.

mpusz:
> There are several places to have the same flow in the file except the place you patched. Should we fix them too?

I didn't fix others because I did not know about them. I just run into this particular one and fixed it. But yes, all the bugs we are aware of should be addressed :wink:

>The "start" is the argument passed into. So if we assign a value to it, I think we need another variable.

This depends on the coding conventions. It is fine to override function arguments and lots of code is doing it (i.e. see reference implementations here: https://en.cppreference.com/w/cpp/algorithm/find). However, if @chriskohlhoff would like me to create a dedicated variable for this one than I will do it.

PR #476, opened 2020-04-06
Adjust ASIO_CONCEPT to work for C++20 (fixes #474)

### LINKED ISSUE #474: C++20 concepts breaks Asio unconditionally
Hi Chris, 

Current Asio (and Boost.Asio) supports Concepts.TS when detected in the compiler.

Unfortunately, C++20 has changed the syntax of concept definition and the rules around constraints. As a result, any recent compiler (clang 10, gcc9 with `-fconcepts`, MSVC `/std:latest`) will fail to compile an asio program.

Boost 1.73 is nearing its release date and I am concerned that if asio is not patched to detect C++20 rather than concepts.ts (probably not a trivial task when considering the entire compiler landscape), as use of newer compilers and the C++20 standard becomes widespread, both Boost and Asio will suffer undue reputation damage.

I'm very happy to contribute time to help provide a patch if you're happy to accept a PR.

Please by all means contact me upon receipt.

[hodges.r@gmail.com](mailto:hodges.r@gmail.com)

BRevzin:


chriskohlhoff:
This was fixed in 00157db09efc9ce65e01ea37346a06b3d82d4f65.

PR #549, opened 2020-09-20
Pass MSG_NOSIGNAL to sendmsg on OpenBSD

nickg:
This avoids Asio applications exiting with unexpected SIGPIPEs on
OpenBSD. Note that FreeBSD has the SO_NOSIGPIPE socket option so doesn't
need this.

### DIFF
--- a/asio/include/asio/detail/impl/socket_ops.ipp
+++ b/asio/include/asio/detail/impl/socket_ops.ipp
@@ -1387,9 +1387,9 @@ signed_size_type send(socket_type s, const buf* bufs, size_t count,
   msghdr msg = msghdr();
   msg.msg_iov = const_cast<buf*>(bufs);
   msg.msg_iovlen = static_cast<int>(count);
-#if defined(__linux__)
+#if defined(__linux__) || defined(__OpenBSD__)
   flags |= MSG_NOSIGNAL;
-#endif // defined(__linux__)
+#endif // defined(__linux__) || defined(__OpenBSD__)
   signed_size_type result = ::sendmsg(s, &msg, flags);
   get_last_error(ec, result < 0);
   return result;
@@ -1418,9 +1418,9 @@ signed_size_type send1(socket_type s, const void* data, size_t size,
   ec.assign(0, ec.category());
   return bytes_transferred;
 #else // defined(ASIO_WINDOWS) || defined(__CYGWIN__)
-#if defined(__linux__)
+#if defined(__linux__) || defined(__OpenBSD__)
   flags |= MSG_NOSIGNAL;
-#endif // defined(__linux__)
+#endif // defined(__linux__) || defined(__OpenBSD__)
   signed_size_type result = ::send(s,
       static_cast<const char*>(data), size, flags);
   get_last_error(ec, result < 0);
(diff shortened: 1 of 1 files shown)


namtsui:
This is a good thing to consider. I also came across this while getting a unit test (test_url_seed/ url_seed_ssl) to pass on OpenBSD for libtorrent-rasterbar.

I had used a reduced version of this patch that doesn't cover all the sendmsg possibilities. This did allow the test to pass.

```
$OpenBSD$

Index: boost/asio/detail/impl/socket_ops.ipp
--- boost/asio/detail/impl/socket_ops.ipp.orig
+++ boost/asio/detail/impl/socket_ops.ipp
@@ -1178,7 +1178,7 @@ signed_size_type send(socket_type s, const buf* bufs,
   msghdr msg = msghdr();
   msg.msg_iov = const_cast<buf*>(bufs);
   msg.msg_iovlen = static_cast<int>(count);
-#if defined(__linux__)
+#if defined(__linux__) || defined(__OpenBSD__)
   flags |= MSG_NOSIGNAL;
 #endif // defined(__linux__)
   signed_size_type result = error_wrapper(::sendmsg(s, &msg, flags), ec);
```

@arvidn from libtorrent-rasterbar had suggested that this submitted as a PR to boost asio. I have not done so. Thank you for submitting this.

explanation: https://github.com/arvidn/libtorrent/issues/4211#issuecomment-570172833
feedback from @arvidn: https://github.com/arvidn/libtorrent/issues/4211#issuecomment-570176356

brad0:
> This avoids Asio applications exiting with unexpected SIGPIPEs on
> OpenBSD. Note that FreeBSD has the SO_NOSIGPIPE socket option so doesn't
> need this.

I see NetBSD also has the socket option. Though the Asio code does not utilize it on
NetBSD at the moment.

namtsui:
Based on these hints from the OpenBSD generic porting tips and Sascha
Wildner from DragonflyBSD, `#ifdef SO_NOSIGPIPE` and `#ifdef MSG_NOSIGNAL`
could be used instead to test directly for the features.

https://www.openbsd.org/faq/ports/guide.html#PortsGeneric

> __OpenBSD__ should be used sparingly, if at all. Constructs that look like
> 
> #if defined(__NetBSD__) || defined(__FreeBSD__)
> 
> are often inappropriate. Don't add blindly __OpenBSD__ to it. Instead, try to figure out what's going on, and what actual feature is needed.

https://www.dragonflybsd.org/mailarchive/users/2011-03/msg00028.html

> A saner way would be to replace those "#ifdef BSD" with "#ifdef SO_NOSIGPIPE" in order to check if the system allows turning off SIGPIPE for a socket. This is how for example BIND does it.

After @brad0's mention of NetBSD, I went through and checked all the
BSDs. NetBSD and FreeBSD have SO_NOSIGPIPE. DragonflyBSD and OpenBSD do not have SO_NOSIGPIPE.

**SO_NOSIGPIPE**:
netbsd: https://man.netbsd.org/man.netbsd.org/setsockopt.2
freebsd: https://www.freebsd.org/cgi/man.cgi?query=setsockopt&apropos=0&sektion=2&manpath=FreeBSD+12.1-RELEASE+and+Ports&arch=default&format=html

dragonflybsd: https://man.dragonflybsd.org/?command=setsockopt&section=ANY
openbsd: https://man.openbsd.org/setsockopt.2

netbsd, freebsd, dragonflybsd and openbsd all have MSG_NOSIGNAL.

**MSG_NOSIGNAL**:
netbsd: https://man.netbsd.org/sendmsg.2
freebsd: https://www.freebsd.org/cgi/man.cgi?query=sendmsg&apropos=0&sektion=2&manpath=FreeBSD+12.1-RELEASE+and+Ports&arch=default&format=html
dragonflybsd: https://man.dragonflybsd.org/?command=sendmsg&section=ANY
openbsd: https://man.openbsd.org/sendmsg.2

nickg:
`MSG_NOSIGNAL` was standardised in POSIX.1-2008 so I suppose the proper fix is to enable it based on `_POSIX_VERSION`. I've done this separately at nickg@e19aa862022 and can update the pull request here if it's acceptable.

pgroke-dt:
Fixing this for Solaris and AIX would also be highly welcome.

chriskohlhoff:

PR #546, opened 2020-09-14
Improve build docs discoverability and add some missing macro documentation (like ASIO_NO_DEPRECATED )

audaki:

PR #562, opened 2020-10-24
Add ASIO_NO_DEPRECATED to doc

kiwixz:


mabrarov:
This pull request duplicates #546 (which I don't no the reason is still not merged).

kiwixz:
Good catch, a quick search did not find it...

audaki:
I've added that specific flag into the title for better searchability until it is merged

PR #566, opened 2020-10-29
Allow using libpthread on Windows

petrutlucian94:
In some cases, using libpthread on Windows is desired, for example
when porting software that's already using libpthread, avoiding
mixing the two libraries, which can be troublesome.

This change will allow using libthread when targetting Windows and
ASIO_HAS_PTHREADS is set.

Boost change: https://github.com/boostorg/asio/pull/366


chriskohlhoff:

PR #584, opened 2020-12-11
Fix duplicate symbols of asio::dummy_return<void> for MSVC

ewienik:
When compiling with Visual Studio there shows LNK2005 (symbol already defined) for asio::dummy_result<void>

PR #587, opened 2020-12-17
Mark template as inline to avoid duplicate symbol errors

gjasny:
```
LINK : bin\test_asyncly_d.exe not found or not built by the last incremental link; performing full link
ObservableTest.cpp.obj : error LNK2005: "void __cdecl boost::asio::dummy_return<void>(void)" (??$dummy_return@X@asio@boost@@YAXXZ) already defined in CoroutineTest.cpp.obj
bin\test_asyncly_d.exe : fatal error LNK1169: one or more multiply defined symbols found
```

### DIFF
--- a/asio/include/asio/impl/use_awaitable.hpp
+++ b/asio/include/asio/impl/use_awaitable.hpp
@@ -236,7 +236,7 @@ T dummy_return()
 }
 
 template <>
-void dummy_return()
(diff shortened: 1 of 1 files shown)


gjasny:
dumplicate of #584. Sorry I only checked boostorg/asio for existing reports.

PR #590, opened 2020-12-28
FIX: host and service name string construction from uninitialized cha…

domsoll:
FIX: host and service name string construction from uninitialized char array at result_type::create call when getnameinfo failed


TheStormN (approved):

TheStormN:
Hey @chriskohlhoff 

This is a really simple fix which can avoid reading garbage data or even reading outside of the buffer. Could you consider merging it?

chriskohlhoff:
Likely addressed by commit 5659422c5, released in asio 1.19.0.

PR #591, opened 2020-12-28
Add this_thread null check for scheduler compensating_work_started routine

cpp11nullptr:
Add this_thread null check for scheduler compensating_work_started routine to be aligned with other routines.

### DIFF
--- a/asio/include/asio/detail/impl/scheduler.ipp
+++ b/asio/include/asio/detail/impl/scheduler.ipp
@@ -320,8 +320,8 @@ void scheduler::restart()
 
 void scheduler::compensating_work_started()
 {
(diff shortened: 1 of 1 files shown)


TheStormN:
I would suggest splitting the check to 2 lines. Most static analyzers will complain for code like this.

chriskohlhoff:
Sorry, this change is not correct. The `compensating_work_started` function is *only* called from inside the scheduler where we must have a valid, non-null call stack. That it's null may indicate that something is seriously wrong with the way the program is built. (A build issue with shared libraries perhaps?) Suppressing it like this will instead introduce a subtle, hard-to-debug work counting problem.

PR #775, opened 2021-01-06
Fix outstanding_work.tracked executor move assignment.

palebedev:
Move assignments for io_context and thread pool executors are missing the logic present in copy assignments that finishes work on old contexts.
This causes the following example to hang:

```
#include <boost/asio/execution/outstanding_work.hpp>
#include <boost/asio/prefer.hpp>
#include <boost/asio/static_thread_pool.hpp>

int main()
{
    asio::static_thread_pool tp1{1},tp2{1};
    {
        auto work = asio::prefer(tp1.get_executor(),
            asio::execution::outstanding_work.tracked);
        work = asio::prefer(tp2.get_executor(),
            asio::execution::outstanding_work.tracked);
    }
    tp1.join();
    tp2.join();
}
```

### DIFF
--- a/asio/include/asio/impl/io_context.hpp
+++ b/asio/include/asio/impl/io_context.hpp
@@ -244,11 +244,16 @@ io_context::basic_executor_type<Allocator, Bits>::operator=(
 {
   if (this != &other)
   {
+    io_context* old_io_context = io_context_;
     io_context_ = other.io_context_;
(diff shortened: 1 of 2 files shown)


chriskohlhoff:
Fixed in 00d6e4ccca14e96808358d728ed032a909b18fc7.

PR #817, opened 2021-05-05
Make allocators respect alignment requirements.

de-vri-es:
This PR makes `recycling_allocator` and `hook_allocator` respect alignment requirement. This is important when completion handlers contain over-aligned members, for example when captured in a lambda.

This is implemented with the C++17 `::operator new(size, align)` if available, and falls back to Boost Align. If Boost Align is also not available, the alignment is silently ignored as before.

I briefly considered implementing a fallback aligned allocation by allocating too much memory to put additional bookkeeping right before the returned pointer. However, I don't think that is worth the added complexity. I think it makes more sense for a user to enable C++17 or Boost if they require proper alignment.

I ran into this in practise when posting a lambda with over-aligned captured Eigen objects using this code:
```c++
#include <asio.hpp>
#include <Eigen/Geometry>
#include <iostream>

void __attribute__((noinline)) foo(
	Eigen::Isometry3d const & a
) {
	std::cout << "foo() a % 32: " << (std::size_t(&a) % 32) << "\n";
};

int main() {
	asio::io_context io_context;
	std::cout << "alignof(Eigen::Isometry3d): " << alignof(Eigen::Isometry3d) << "\n";

	auto b = Eigen::Isometry3d::Identity();
	asio::post(io_context, [b] () {
		foo(b);
	});

	io_context.run();
}
```

When compiled with `g++ -std=c++17 -pthread -mavx` without this patch, the code above produces the following output:
```
alignof(Eigen::Isometry3d): 32
a.out: /usr/include/eigen3/Eigen/src/Core/DenseStorage.h:128: Eigen::internal::plain_array<T, Size, MatrixOrArrayOptions, 32>::plain_array() [with T = double; int Size = 16; int MatrixOrArrayOptions = 0]: Assertion `(internal::UIntPtr(eigen_unaligned_array_assert_workaround_gcc47(array)) & (31)) == 0 && "this assertion is explained here: " "http://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAssert.html" " **** READ THIS WEB PAGE !!! ****"' failed.
```

With the PR, the compiled binary works as expected if C++17 is available, or if C++11 and Boost Align are available.

### DIFF
--- a/asio/include/asio/detail/aligned_alloc.hpp
+++ b/asio/include/asio/detail/aligned_alloc.hpp
@@ -1,0 +1,55 @@
+//
+// detail/aligned_alloc.hpp
+// ~~~~~~~~~~~~~~~~
+//
+// Copyright (c) 2021 Maarten de Vries (maarten at de-vri dot es)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+
+#ifndef ASIO_DETAIL_ALIGNED_ALLOC_HPP
+#define ASIO_DETAIL_ALIGNED_ALLOC_HPP
+
+#if defined(_MSC_VER) && (_MSC_VER >= 1200)
+# pragma once
+#endif // defined(_MSC_VER) && (_MSC_VER >= 1200)
+
+#include "asio/detail/config.hpp"
+
+#if !defined(ASIO_HAS_ALIGNED_NEW) \
+  && defined(ASIO_HAS_BOOST_ALIGN) \
+  && defined(ASIO_HAS_ALIGNOF)
+# include <boost/align/aligned_alloc.hpp>
+#else
+# include <new>
+#endif
+
+namespace asio {
+namespace detail {
+
+inline void* aligned_alloc(std::size_t align, std::size_t size) {
+#if defined(ASIO_HAS_ALIGNED_NEW) && defined(ASIO_HAS_ALIGNOF)
+  return ::operator new(size, std::align_val_t(align));
+#elif defined(ASIO_HAS_BOOST_ALIGN) && defined(ASIO_HAS_ALIGNOF)
+  return boost::alignment::aligned_alloc(align, size);
+#else
+  (void) align;
+  return ::operator new(size);
+#endif
+}
+
+inline void aligned_free(void* ptr) {
+#if !defined(ASIO_HAS_ALIGNED_NEW) \
+  && defined(ASIO_HAS_BOOST_ALIGN) \
+  && defined(ASIO_HAS_ALIGNOF)
+  boost::alignment::aligned_free(ptr);
+#else
+  ::operator delete(ptr);
+#endif
+}
+
+} // namespace detail
+} // namespace asio
+
+#endif // ASIO_DETAIL_ALIGNED_ALLOC_HPP
--- a/asio/include/asio/detail/config.hpp
+++ b/asio/include/asio/detail/config.hpp
@@ -26,6 +26,7 @@
 
 // boostify: non-boost code ends here
 #if defined(ASIO_STANDALONE)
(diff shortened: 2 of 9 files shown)


asio/include/asio/detail/aligned_alloc.hpp
```diff
@@ -0,0 +1,54 @@
...
+namespace detail {
+
+inline void* aligned_alloc(std::size_t align, std::size_t size) {
+#if defined(ASIO_HAS_ALIGNED_NEW) && defined(ASIO_HAS_ALIGNOF)
+  return ::operator new(size, std::align_val_t(align));
+#elif defined(ASIO_HAS_BOOST_ALIGN) && defined(ASIO_HAS_ALIGNOF)
+  return boost::alignment::aligned_alloc(align, size);
+#else
+  return ::operator new(size);
+#endif
```
de-vri-es, line 38: I purposefully ignore the `align` argument when `alignof` is not available, since it will just be `1`, and I'm not sure if that would cause some allocators to return unaligned allocations.

If they do, that would break alignment even of types that are not over-aligned.

asio/include/asio/detail/config.hpp
```diff
@@ -559,6 +560,32 @@
...
+# if !defined(ASIO_DISABLE_ALIGNOF)
+#  if (__cplusplus >= 201103)
+#   define ASIO_HAS_ALIGNOF 1
+#  endif // (__cplusplus >= 201103)
+# endif // !defined(ASIO_DISABLE_ALIGNOF)
+#endif // !defined(ASIO_HAS_ALIGNOF)
+
+#if defined(ASIO_HAS_ALIGNOF)
+# define ASIO_ALIGNOF(T) alignof(T)
+# define ASIO_DEFAULT_ALIGN alignof(std::max_align_t)
```
de-vri-es, line 574: Note that `max_align_t` gives the maximum alignment of scalar types and not the maximum possible alignment. Dynamic allocations should already be aligned to that value, so this shouldn't change the alignment of non over-aligned types.

asio/include/asio/detail/aligned_alloc.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+inline void* aligned_alloc(std::size_t align, std::size_t size) {
+#if defined(ASIO_HAS_ALIGNED_NEW) && defined(ASIO_HAS_ALIGNOF)
+  return ::operator new(size, std::align_val_t(align));
+#elif defined(ASIO_HAS_BOOST_ALIGN) && defined(ASIO_HAS_ALIGNOF)
+  return boost::alignment::aligned_alloc(align, size);
+#else
+  (void) align;
+  return ::operator new(size);
+#endif
```
de-vri-es, line 39: I purposefully ignore the `align` argument when `alignof` is not available, since it will just be `1`, and I'm not sure if that would cause some allocators to return unaligned allocations.

If they do, that would break alignment even of types that are not over-aligned.

de-vri-es:
There seems to be a heap corruption on windows with mingw32 in the buffered stream test. I'm guessing this is caused by memory being allocated by an `operator new()` or regular `alloc()`, and then freed by `aligned_free`, or perhaps the other way around.

I'll dig deeper to figure out if that assumption is correct and fix it.

de-vri-es:
After spending most day trying to figure out how to get anything done on Windows ( D: ), I finally managed to get a stack trace for the failing test. That directly pointed the finger at the destructor of `thread_info_base`, which was indeed still calling raw `operator delete`.

I've replaced it with a call to the new `aligned_free` function, and I'm optimistic that the unit test will now pass on all platforms (they did locally).

de-vri-es:
Polite ping @chriskohlhoff. Do you have some time to review this? This can currently cause UB :]

chriskohlhoff:
A tweaked version has been committed to master.

de-vri-es:
Awesome! Thank you for the review and of course the great library :)

PR #848, opened 2021-06-08
Fix variable reference in sockatmark()

StephenCWills:
This branch of the preprocessor logic is referencing a variable from the other branch, so it doesn't exist. I assume this was supposed to be checking for errors on `value` instead of `result`.


StephenCWills:
@chriskohlhoff Since it seems to me that this was effectively merged via f79d3dc, I'm closing the PR. If there's a reason you were keeping this open, feel free to reopen.

PR #849, opened 2021-06-08
Reorder field members in basic_resolver_entry

StephenCWills:
I've been working with a platform that defines the `sockaddr` struct using a flexible array member like so.

```C
struct sockaddr {
    sa_family_t sa_family;
    char        sa_data[];
};
```

This was causing problems for `basic_resolver_entry` because the `endpoint_` field isn't placed at the end of the class definition. Seeing as how this isn't necessarily a supported platform, I wasn't sure if you'd really want to accept this change. Having said that, the change is fairly benign, and I believe this may be a valid definition for the `sockaddr` struct in general so I also thought you might find this change to be worth considering.


AJenbo:
@StephenCWills should this change be enough to set networking running on 3DS?

StephenCWills:
> should this change be enough to set networking running on 3DS?

We need this one as well as #848 and https://github.com/StephenCWills/asio/commit/f054d98c32a1bab5846d975f238c70c88658110a.

PR #854, opened 2021-06-20
Added promise & coro.

klemens-morgenstern:
Coros will take a bit longer to port I fear.


microcai:
when try to use  promise<>::race(), I get compile error:

'promise' has been explicitly marked deleted here
 promise(const promise& ) = delete;

note: passing argument to parameter 'ps' here
  static auto race(Executor1 exec, Ps ... ps)

chriskohlhoff:
Likely addressed by commit 7e3d99643, released in asio 1.19.0.

PR #862, opened 2021-07-03
yield input fix.

klemens-morgenstern:


klemens-morgenstern:
@chriskohlhoff Am I right to think you merged this into master already?

PR #871, opened 2021-07-12
Different fix for strand "move after use" bug that should not make extra copies

### LINKED ISSUE #866: Strand destroys its executor and causes "use after move" crash when require+prefer return original executor
`strand_executor_service` seems to have undefined behavior when working with executors with specific properties.

Destructor of `on_invoker_exit` checks if there are pending handlers in strand's queue, and if yes, it calls `execution::execute` with 2 parameter - rvalue-ref to `this_` and  executor. The executor is a result of `require` and `prefer` applied to `this_->executor_`.
Depending on executor's implementation and properties it may happen that applying properties to `this_->executor_` will return a reference to `this_->executor_` itself. If I understand it correctly, this happens if the executor already satisfies required properties, and either already satisfies or doesn't support optional properties (requested using `prefer`). In this case call to `execute` essentially becomes:
`execution::execute(this_->executor_, std::move(this_))`
where first parameter is lvalue-ref.
See https://github.com/chriskohlhoff/asio/blob/asio-1-18-2/asio/include/asio/detail/impl/strand_executor_service.hpp#L108

Then `execution::execute` does
`return ASIO_MOVE_CAST(T)(t).execute(ASIO_MOVE_CAST(F)(f));`
which is undefined behavior since order of moves is undefined.
In my case `ASIO_MOVE_CAST(T)(t)` happens first, so `this_->executor_` is getting moved and `this_` is left with invalid executor. On next iteration it crashes trying to use invalid executor. If it was the other way around it would probably crash sooner.
See https://github.com/chriskohlhoff/asio/blob/asio-1-18-2/asio/include/asio/execution/execute.hpp#L204

Here is a simplified code that demonstrates the problem (I use gcc 7.4.0 on Linux and boost 1.76):
```
#include <cassert>

#include <boost/asio/io_context.hpp>
#include <boost/asio/spawn.hpp>
#include <boost/asio/strand.hpp>
#include <gtest/gtest.h>

using namespace boost::asio;

class Executor;

class Context : public boost::asio::execution_context {
 public:
  using executor_type = Executor;

 Executor get_executor();

    void enqueue(std::function<void()> f) {
	tasks_.push_back(std::move(f));
    }

    std::vector<std::function<void()>> tasks_;

    void run() {
	for (auto& t: tasks_) {
	    t();
	}
	tasks_.clear();
    }
};

class Executor {
 public:
  Executor(Context* c) : context_(c) {}
  Executor(Executor&& other) noexcept
      : context_(other.context_) {
      other.moved_out_ = true;
  };
  Executor& operator=(Executor&& other) noexcept  {
      context_ = other.context_;
      other.moved_out_ = true;
      return *this;
  };
  Executor(const Executor&) noexcept = default;
  Executor& operator=(const Executor&) noexcept = default;
  bool operator==(const Executor& other) const noexcept {
    return context_ == other.context_;
  }
  bool operator!=(const Executor& other) const noexcept {
    return !(*this == other);
  }

  boost::asio::execution_context& query(boost::asio::execution::context_t) const
      noexcept {
    return *context_;
  }

  static constexpr boost::asio::execution::blocking_t::never_t query(
      boost::asio::execution::blocking_t) {
    return boost::asio::execution::blocking.never;
  }

  Context& context() const noexcept { return *context_; }

  template <class F>
  void execute(F f) const {
      assert(!moved_out_);
      context_->enqueue(f);
  }

 private:
    Context* context_;
    bool moved_out_ = false;
};

Executor Context::get_executor() {
    return Executor(this);
}

int main() {
    Context context;
    auto strand = boost::asio::make_strand(context.get_executor());
    execution::execute(strand, []{});
    execution::execute(strand, []{});
    context.run();
}
```

Stack trace and variable addresses printed by gdb that show that `t` (executor) is a part of `f` (callable object) in `execute`:

```
void Executor::execute(F) const [with F = boost::asio::detail::strand_executor_service::invoker<const Executor, void>]: Assertion `!moved_out_' failed.

Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff7264801 in __GI_abort () at abort.c:79
#2  0x00007ffff725439a in __assert_fail_base (fmt=0x7ffff73db7d8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x5555555623b9 "!moved_out_", 
    file=file@entry=0x555555562368 "/home/antonperkov/workspace/voyager/onboard/common/node/src/asio_strand_test.cpp", line=line@entry=67, 
    function=function@entry=0x5555555623e0 <void Executor::execute<boost::asio::detail::strand_executor_service::invoker<Executor const, void> >(boost::asio::detail::strand_executor_service::invoker<Executor const, void>) const::__PRETTY_FUNCTION__> "void Executor::execute(F) const [with F = boost::asio::detail::strand_executor_service::invoker<const Executor, void>]") at assert.c:92
#3  0x00007ffff7254412 in __GI___assert_fail (assertion=0x5555555623b9 "!moved_out_", 
    file=0x555555562368 "/home/antonperkov/workspace/voyager/onboard/common/node/src/asio_strand_test.cpp", line=67, 
    function=0x5555555623e0 <void Executor::execute<boost::asio::detail::strand_executor_service::invoker<Executor const, void> >(boost::asio::detail::strand_executor_service::invoker<Executor const, void>) const::__PRETTY_FUNCTION__> "void Executor::execute(F) const [with F = boost::asio::detail::strand_executor_service::invoker<const Executor, void>]") at assert.c:101
#4  0x0000555555560a5e in Executor::execute<boost::asio::detail::strand_executor_service::invoker<Executor const, void> > (this=0x55555577e6d0, f=...)
    at /home/antonperkov/workspace/voyager/onboard/common/node/src/asio_strand_test.cpp:67
#5  0x00005555555615e2 in asio_execution_execute_fn::impl::operator()<Executor&, boost::asio::detail::strand_executor_service::invoker<Executor const, void> > (
    this=0x555555562d97 <asio_execution_execute_fn::static_instance<asio_execution_execute_fn::impl>::instance>, t=..., f=...)
    at /home/antonperkov/workspace/boost_1_76_0/boost/asio/execution/execute.hpp:208
#6  0x00005555555614fc in boost::asio::detail::strand_executor_service::invoker<Executor const, void>::on_invoker_exit::~on_invoker_exit (this=0x7fffffffe1e0, 
    __in_chrg=<optimized out>) at /home/antonperkov/workspace/boost_1_76_0/boost/asio/detail/impl/strand_executor_service.hpp:109
#7  0x0000555555561324 in boost::asio::detail::strand_executor_service::invoker<Executor const, void>::operator() (this=0x55555577e6c0)
    at /home/antonperkov/workspace/boost_1_76_0/boost/asio/detail/impl/strand_executor_service.hpp:122
#8  0x000055555556114d in std::_Function_handler<void (), boost::asio::detail::strand_executor_service::invoker<Executor const, void> >::_M_invoke(std::_Any_data const&) (
    __functor=...) at /usr/include/c++/7/bits/std_function.h:316
#9  0x000055555555d7c0 in std::function<void ()>::operator()() const (this=0x55555577e6f0) at /usr/include/c++/7/bits/std_function.h:706
#10 0x000055555555c0b8 in Context::run (this=0x7fffffffe2b0) at /home/antonperkov/workspace/voyager/onboard/common/node/src/asio_strand_test.cpp:26
#11 0x00005555555579da in main () at /home/antonperkov/workspace/voyager/onboard/common/node/src/asio_strand_test.cpp:95
(gdb) fr 5
#5  0x00005555555615e2 in asio_execution_execute_fn::impl::operator()<Executor&, boost::asio::detail::strand_executor_service::invoker<Executor const, void> > (
    this=0x555555562d97 <asio_execution_execute_fn::static_instance<asio_execution_execute_fn::impl>::instance>, t=..., f=...)
    at /home/antonperkov/workspace/boost_1_76_0/boost/asio/execution/execute.hpp:208
208	    return BOOST_ASIO_MOVE_CAST(T)(t).execute(BOOST_ASIO_MOVE_CAST(F)(f));
(gdb) p &t
$1 = (Executor *) 0x55555577e6d0
(gdb) p &f->executor_
$2 = (boost::asio::detail::strand_executor_service::invoker<Executor const, void>::executor_type *) 0x55555577e6d0
```

I think this happens with my executor because my executor doesn't support encapsulating allocators. In my real code (different from simplified example above) I don't need it because I do only one allocation, and that allocation must be done using  API provided by a third-party library (my execution context is a wrapper around the library that actually executes work). Executor of `boost::asio::io_context` supports custom allocators. Therefore applying `recycling_allocator<void>` to executor returns an executor of a different type (maybe unless it has been already applied before), so returns a different object and downstream code doesn't invalidate original executor.

Is this a bug?
It seems to me like yes, because I don't see a requirement that every executor should support custom allocators.

If this is a bug, what exactly is wrong here?

One option is that `boost::asio::prefer` and `boost::asio::require` should never return lvalue reference to executor. Because it's very confusing and error prone that depending on executor type and requested properties, result can be either reference to the same object or a new object.

Another option is that `strand_executor_service::invoker::on_invoker_exit` should be fixed to always make a copy of `this_->executor_`. I've made this fix in my local copy of asio and it solved the problem.

There is also the option that all executors must support allocators. In this case I think it deserves a `static_assert` or some sort of compile-time check (although this requirement doesn't seem reasonable to me).

moskvanaft:
Another way to fix https://github.com/chriskohlhoff/asio/issues/866

### DIFF
--- a/asio/include/asio/detail/impl/strand_executor_service.hpp
+++ b/asio/include/asio/detail/impl/strand_executor_service.hpp
@@ -120,8 +120,8 @@ public:
               execution::blocking.never),
             execution::allocator(allocator));
 
-        execution::execute(BOOST_ASIO_MOVE_CAST(new_executor_type)(ex),
-            BOOST_ASIO_MOVE_CAST(invoker)(*this_));
+        execution::execute(ASIO_MOVE_CAST(new_executor_type)(ex),
+            ASIO_MOVE_CAST(invoker)(*this_));
       }
     }
   };
--- a/asio/include/asio/detail/impl/strand_executor_service.hpp
+++ b/asio/include/asio/detail/impl/strand_executor_service.hpp
@@ -107,18 +107,18 @@ public:
         recycling_allocator<void> allocator;
 	typedef typename decay<
             typename prefer_result<
-	      typename require_result<
-		executor_type,
-		execution::blocking_t::never_t
-	      >::type,
+              typename require_result<
+                executor_type,
+                execution::blocking_t::never_t
+              >::type,
               execution::allocator_t<recycling_allocator<void>>
             >::type
           >::type new_executor_type;
 
-	new_executor_type ex = boost::asio::prefer(
-	    boost::asio::require(this_->executor_,
+        new_executor_type ex = boost::asio::prefer(
+            boost::asio::require(this_->executor_,
               execution::blocking.never),
-	    execution::allocator(allocator));
+            execution::allocator(allocator));
 
         execution::execute(BOOST_ASIO_MOVE_CAST(new_executor_type)(ex),
             BOOST_ASIO_MOVE_CAST(invoker)(*this_));
--- a/asio/include/asio/detail/impl/strand_executor_service.hpp
+++ b/asio/include/asio/detail/impl/strand_executor_service.hpp
@@ -115,8 +115,8 @@ public:
             >::type
           >::type new_executor_type;
 
-        new_executor_type ex = boost::asio::prefer(
-            boost::asio::require(this_->executor_,
+        new_executor_type ex = asio::prefer(
+            asio::require(this_->executor_,
               execution::blocking.never),
             execution::allocator(allocator));
 
--- a/asio/include/asio/detail/impl/strand_executor_service.hpp
+++ b/asio/include/asio/detail/impl/strand_executor_service.hpp
@@ -105,13 +105,23 @@ public:
       if (push_waiting_to_ready(this_->impl_))
       {
         recycling_allocator<void> allocator;
-        executor_type ex = this_->executor_;
-        execution::execute(
-            asio::prefer(
-              asio::require(ex,
-                execution::blocking.never),
-            execution::allocator(allocator)),
-            ASIO_MOVE_CAST(invoker)(*this_));
+	typedef typename decay<
+            typename prefer_result<
+	      typename require_result<
+		executor_type,
+		execution::blocking_t::never_t
+	      >::type,
+              execution::allocator_t<recycling_allocator<void>>
+            >::type
+          >::type new_executor_type;
+
+	new_executor_type ex = boost::asio::prefer(
+	    boost::asio::require(this_->executor_,
+              execution::blocking.never),
+	    execution::allocator(allocator));
+
+        execution::execute(BOOST_ASIO_MOVE_CAST(new_executor_type)(ex),
+            BOOST_ASIO_MOVE_CAST(invoker)(*this_));
       }
     }
   };


chriskohlhoff:
Likely already fixed / superseded by commit 92971f12b, released in asio 1.19.1.

PR #872, opened 2021-07-12
Fix null dereference in awaitable_thread destructor.

FarisElghlan:
If the variable `bottom_of_stack_` is moved into the lambda capture before the expression `bottom_of_stack_.frame_->u_.executor_` is evaluated, a read access violation will occur.

I think this problem depends on the compiler, as the order in which function parameters are evaluated is unspecified. I observed this issue while compiling with MSVC.


chriskohlhoff:
Thanks for the report. Fixed in cafb11a551ed4f4011ec1cb128be612912d8cfd6, which takes a slightly different approach.

PR #879, opened 2021-07-28
Fix posix::basic_stream_descriptor move operations

RedDwarf69:
They were only working when using the default executor.


chriskohlhoff:
Merged in c243c682458aacd5e2c21cb75f4b0cbad19c766c.

PR #905, opened 2021-09-17
Fix the initialisation order of members

BenHetherington:
Reorder the constructor list in `win_iocp_socket_accept_op`, so that `proxy_op` comes before `cancel_requested`, matching the order in which these members are declared (and the order in which they are initialised).

This addresses a warning that members are constructed in a different order to what the constructor implies. (I appreciate that you may not be able to guarantee that things build with all warnings in all compilers, but this is at least a simple one to resolve!)


BenHetherington:
This appears to have been cherry-picked as cd68912b7eb07f24e501e8cbff832b61ec292eea.

PR #909, opened 2021-09-29
Mark ~cancellation_handler_base() virtual

Osse:
This avoids C4265 on MSVC.


klemens-morgenstern:
That's an unnecessary warning of your compiler though, not an issue in the code.

viccpp:
The compiler must be fixed not the code. This destructor is protected

PR #910, opened 2021-09-30
Enable ASIO_HAS_STD_INVOKE_RESULT with libc++13

jcelerier:
In c++20 mode std::result_of has been retired as per the standard, causing compile errors.


chriskohlhoff:
Likely addressed by commit b3a03038c, released in asio 1.21.0.

PR #914, opened 2021-10-12
Added notify to timers.

klemens-morgenstern:


klemens-morgenstern:
Closing due to rejection by Chris.

PR #924, opened 2021-10-22
use faster ctor for network_v4 in canonical()

SaltfishAmi:
`network_v6::canonical()` already uses prefix_length().


SaltfishAmi:
fixed in boostorg/asio@46c729aa286f1577421b05d165568b60de0cd509

PR #936, opened 2021-11-13
Fix any_io_executor with ASIO_USE_TS_EXECUTOR_AS_DEFAULT

tanzislam:
Commit bae9bcf has overlooked the case of `any_io_executor` being a typedef to `executor` (when `ASIO_USE_TS_EXECUTOR_AS_DEFAULT` is defined) instead of a proper class, resulting in compilation errors.


tanzislam:
Merged in commit 689cf69. Thanks!

PR #938, opened 2021-11-14
Added && async_resume overload.

klemens-morgenstern:
Currently this is UB:

```cpp
coro<void> task(asio::any_io_executor exec);

int main()
{
   asio::io_context ctx;
   
   task(ctx.get_executor()).async_resume(asio::detached);l
   
   ctx.run();
   
   return 0;
}
```

This is because the coro gets destructed while / before being resumed. This PR changes the coro interface, so that calling `async_resume` on a coro will be equivalent to `co_spawn` , i.e. steal the coro handle and have the coro live in the executor.

### DIFF
--- a/asio/include/asio/experimental/co_spawn.hpp
+++ b/asio/include/asio/experimental/co_spawn.hpp
@@ -27,25 +27,6 @@
 
 namespace asio {
 namespace experimental {
-namespace detail {
-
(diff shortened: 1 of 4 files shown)


klemens-morgenstern:
Looks like it's been added by other means..

PR #953, opened 2021-12-01
Make awaitable_operators.hpp self contained

andrei-datcu:
It was missing `asio/co_spawn.hpp`


chriskohlhoff:
Likely addressed by commit 5584f7386, released in asio 1.22.0.

PR #954, opened 2021-12-01
Fix awaitable move assignment operator return value

andrei-datcu:


chriskohlhoff:
Likely addressed by commit 607b67c75, released in asio 1.22.0.

PR #958, opened 2021-12-08
Fix buffer size bug in channel example

gummif:


gummif:
Function buffer does this, so not needed.

PR #963, opened 2021-12-14
fix missing ref to send_endpoint in io_uring_socket_recvfrom_op

EvinceMoi:
When using io_uring as backend(with BOOST_ASIO_HAS_IO_URING and BOOST_ASIO_DISABLE_EPOLL defined), the endpoint parameter in async_receive_from of udp_socket is not correctly set due to the missing ref in class io_uring_socket_recvfrom_op.


chriskohlhoff:
Fixed in 1.22.1.

PR #991, opened 2022-02-10
Include concurrency_hint.hpp from io_context.hpp

kiwixz:
The concurrency_hint header should be pulled automatically via io_context.hpp; as it's located under the "detail" include directory I think it was the intended behavior.


chriskohlhoff:
Fixed in 1.22.1.

PR #1003, opened 2022-02-23
Fix promise.hpp: 'all_of': is not a member of 'std'

gummif:
Fix missing including (on VS2019).


chriskohlhoff:
Fixed in 1.22.1.

PR #1021, opened 2022-03-15
Make gcc_x86_fenced_block compatible with -masm=intel

jwt27:
This small tweak allows `gcc -masm=intel` users to compile this header.


chriskohlhoff:
Likely addressed by commit 4de8e6dab, released in asio 1.23.0.

PR #1023, opened 2022-03-16
promise copy executor from first item before moving range

ZerocksX:
This is undefined behavior and crashes on MSVC


ZerocksX:
@klemens-morgenstern Please take a look at this issue

chriskohlhoff:
Likely addressed by commit b2e7ba6c1, released in asio 1.22.2.

Issue #1028, opened 2022-03-28
decay<> ambiguity after using directives


brunom:
Under Visual Studio 2019/2022 including asio after using directives for ::std and ::boost fails to compile:

	#include <boost/optional.hpp>
	#include <type_traits>
	using namespace boost;
	using namespace std;
	#include <boost/asio.hpp>

Errors:

	1>  C:\boost_1_79_0\boost\asio\execution\schedule.hpp(131,1): error C2872: 'decay<`template-type-parameter-1'>': ambiguous symbol
	1>  C:\boost_1_79_0\boost\asio\execution\schedule.hpp(131,1): message : could be 'std::decay<_Ty>'
	1>  C:\boost_1_79_0\boost\asio\execution\schedule.hpp(131,1): message : or       'boost::decay<T>'

I was sent here after reporting the bug to Microsoft: [https://developercommunity.visualstudio.com/t/decay-ambiguous-between-std-and-boost-when-using-n/1613214](https://developercommunity.visualstudio.com/t/decay-ambiguous-between-std-and-boost-when-using-n/1613214)

chriskohlhoff:
This looks like a MSVC compiler bug to me. Please reopen the bug report with them.

https://godbolt.org/z/jrME9bd9d

Suspiciously, the problem only occurs when the so-called ambiguous name is a struct template with definition.

brunom:
@StephanTLavavej could you look at this directly? I don't understand the details at all.

StephanTLavavej:
Very interesting - I initially believed that this was an Asio bug, but @chriskohlhoff's Compiler Explorer example is compelling - I agree that this appears to be an MSVC compiler bug. What's happening here appears to involve a Core Language rule that I've never had to interact with - it looks like [basic.lookup.unqual] says something slightly different than my naive understanding of how using-directives and using-declarations work (which I've gotten away with because our STL implementation avoids ever exercising this rule). (Specifically it looks like N4910 [basic.lookup.unqual]/2.2 makes using-directives a bit weaker than I thought, so the using-declaration here takes priority - but I am not as confident in that reading of the root cause, as I am that Chris's link demonstrates that MSVC is wrong about *something* in this scenario.)

I recommend reporting this to Developer Community with that Godbolt link (using a fresh bug so the previous repro doesn't create a confusing history). I could report it internally, but externally-reported compiler bugs are treated with higher priority (perhaps surprisingly).

brunom:
[...]

PR #1031, opened 2022-04-02
#1028 decay<> ambiguity after using directives

daniel-mdb:
#1028 decay<> ambiguity after using directives

Proposal for fixing the following fragment with MSVC by always referring
to the complete qualifier `::asio::decay`

```
#include <boost/optional.hpp>
#include <type_traits>
using namespace boost;
using namespace std;
#include <boost/asio.hpp>
```


daniel-mdb:
Not needed, the bug lies in the compiler.

PR #1042, opened 2022-04-23
Coropose Experiment (coroutine compose)

klemens-morgenstern:
Heya Chris,

I've written a minimal PoC for a coroutine based async_compose alternative, that is the easiest solution I could come up with. IT does currently only handle callbacks, use_coro & use_awaitable, and now modifiers. But it does work

Let's say we have a composed operation that reads one line and yield the content:

```cpp
template<typename Stream, typename CompletionToken>
auto async_read_line(Stream & stream,  CompletionToken && completion_token,
               asio::async_coropose_tag<void(error_code, std::string)> = {})
               // ^ this tag is where the magic happens, so we can look up the coro type.
      -> asio::async_coropose_t<void(error_code, std::string), CompletionToken>
{
  auto tk = co_await asio::this_coro::token; //yields use_awaitable or use_coro ... or something else in the future
  
  std::string buf;
  auto [sz, ec] = co_await asio::async_read_until(str, asio::dynamic_buffer(buf), asio::as_tuple(tk));
  buf.resize(sz); 
  co_return std::make_tuple{ec, buf};
}
```

The tag allows the `coroutine_traits` to pick up on it being a coroutine. If `completion_token is a lambda, the core promise  becomes the `coropose_awaitable_frame` with the return type being `void`. The interal frame just spawns itself onto the executor of `stream` and does the work.

`async_read_line(stream, function<void(error_code, std::string)>) -> void`.

For a `experimental::coro` and `awaitable` `asio::async_coropose_t<void(error_code, std::string), CompletionToken>` just picks the corresponding type:

```cpp
async_read_line(stream, asio::use_awaitable<Executor>) -> asio::awaitable<std::string>
async_read_line(stream, asio::exp::use_coro<Executor>) -> asio::exp::coro<std::string>
```

Further work would require adding support for modifiers and more tokens, but I think this concept would make writing extensions so much easier; the only easier style would be to just use awaitables everywhere.

### DIFF
--- a/asio/include/asio/awaitable.hpp
+++ b/asio/include/asio/awaitable.hpp
@@ -99,9 +99,14 @@ public:
   }
 
   // Support for co_await keyword.
+
+  template<typename U>
+  static int inherits_awaitable_frame(const detail::awaitable_frame<U, Executor> & );
+
   template <class U>
   void await_suspend(
-      detail::coroutine_handle<detail::awaitable_frame<U, Executor>> h)
+      detail::coroutine_handle<U> h,
+      decltype(inherits_awaitable_frame(std::declval<U>()))= 0)
   {
     frame_->push_frame(&h.promise());
   }
--- a/asio/include/asio/coropose.hpp
+++ b/asio/include/asio/coropose.hpp
@@ -1,0 +1,229 @@
+// Copyright (c) 2021 Klemens D. Morgenstern
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+#ifndef ASIO_COROPOSE_HPP
+#define ASIO_COROPOSE_HPP
+
+
+#if defined(_MSC_VER) && (_MSC_VER >= 1200)
+# pragma once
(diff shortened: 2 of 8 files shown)


klemens-morgenstern:
WIth 843e9dd all completion tokens are supported, defaulting to doing a `co_spawn`.

One minor optimization could be as_single/as_tuple/redicrect_error for `use_awaitable` and `use_coro`, because those result modifiers do not need a co_spawn. The other modifiers (e.g. bind_executor) do always need some for of co_spawning.

klemens-morgenstern:
Closing, I found a better solution:  https://github.com/madmongo1/asio_experiments/pull/4

It seems to be faster than async_compose (tested with clang-14)

```cpp
Naked     composed op took 134700664ns
Recycling composed op took 134154100ns
Naked         coro op took 69197452ns
Recycling     coro op took 72323985ns
```


Awaitable support is still hacky though, so this exentions to the async_resume would be very handy. So I think

[af9870b](https://github.com/chriskohlhoff/asio/pull/1042/commits/af9870ba3f2781086494dc75e9bcd40ab0edd705)
and
[24711c6](https://github.com/chriskohlhoff/asio/pull/1042/commits/24711c6e31d0d16a15a0796f4ce22bba7c8f9e73)

Would be worth adding.

PR #1045, opened 2022-04-24
Fiber

klemens-morgenstern:
Added a fiber, as a `boost.context` based replacement of `boost::spawn`.

### DIFF
--- a/asio/include/asio/fiber.hpp
+++ b/asio/include/asio/fiber.hpp
@@ -1,0 +1,738 @@
+// Copyright (c) 2021 Klemens D. Morgenstern
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+#ifndef ASIO_FIBER_HPP
+#define ASIO_FIBER_HPP
+
+#if defined(_MSC_VER) && (_MSC_VER >= 1200)
+# pragma once
+#endif // defined(_MSC_VER) && (_MSC_VER >= 1200)
+
+#include "asio/detail/config.hpp"
+
+#include <boost/context/fiber.hpp>
+#include <optional>
+
+#include "asio/any_io_executor.hpp"
+#include "asio/associated_cancellation_slot.hpp"
+#include "asio/associated_executor.hpp"
+#include "asio/bind_cancellation_slot.hpp"
+#include "asio/bind_executor.hpp"
+#include "asio/cancellation_state.hpp"
+#include "asio/experimental/deferred.hpp"
+#include "asio/dispatch.hpp"
+#include "asio/is_executor.hpp"
+#include "asio/post.hpp"
+
+#include "asio/detail/push_options.hpp"
+
+namespace asio
+{
+
+namespace detail
+{
+
+struct fiber_handle
+{
+  bool did_suspend{false};
+  boost::context::fiber fiber;
+  std::unique_ptr<fiber_handle> lifetime;
+  cancellation_slot  cancel_slot;
+  cancellation_state cancel_state{cancel_slot};
+
+  fiber_handle(cancellation_slot  cancel_slot) : cancel_slot(std::move(cancel_slot)) {}
+};
+
+}
+
+template<typename Executor = asio::any_io_executor>
+struct basic_fiber_context
+{
+  using executor_type = Executor;
+
+
+  template<typename Executor2>
+  basic_fiber_context(
+          const basic_fiber_context<Executor2>  & other,
+          typename constraint<std::is_convertible_v<Executor2, Executor>>::type = 0) noexcept
+          : handle_(other.handle_), fiber_(other.fiber_), executor_(other.get_executor())
+  {
+  }
+
+  template<typename Executor2>
+  basic_fiber_context(
+          basic_fiber_context<Executor2>  && other,
+          typename constraint<std::is_convertible_v<Executor2, Executor>>::type = 0) noexcept
+          : handle_(other.handle_), fiber_(std::move(other.fiber_)), executor_(std::move(other.get_executor()))
+  {
+  }
+
+  basic_fiber_context(detail::fiber_handle &handle,
+                      boost::context::fiber && fiber,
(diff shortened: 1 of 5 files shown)


asio/src/doc/overview/fiber.qbk
```diff
@@ -0,0 +1,105 @@
...
+
+[section:fiber Stackful Coroutines (boost.context.fiber)]
+
+The [link asio.reference.async_fiber `async_fiber()`] function is a high-level wrapper for
+running stackful coroutines, named fiber in this context.
+It is based on the Boost.Coroutine library.
+The `async_fiber()` function enables programs to implement
+asynchronous logic in a synchronous manner, as shown in the following example:
+
+  asio::async_fiber(my_strand, &do_echom asio::detached);
```
madmongo1, line 17: typo

asio/include/asio/fiber.hpp
```diff
@@ -45,13 +45,17 @@ struct fiber_handle
...
+ * @tparam Executor The executor of the fiber_context.
+*/
 template<typename Executor = asio::any_io_executor>
 struct basic_fiber_context
 {
+  /// The executor of the fiber.
   using executor_type = Executor;
 
-
+  //// Copy construct a basic_fiber_context from anoter context. Can convert the executor type.
```
vinniefalco, line 58: Did you mean to put four slashes?

klemens-morgenstern:
I've added what I call pseudo-coroutines, i.e. `co_await` / `co_return` support to the fibers. It gets picked up by having a `yield_context` as first of second (bc. lambda & member function) argument. The thing just gets run and thus can get completly inlined, i.e. avoid any allocations.

This is nice for the explicit syntax of suspending operations, but it also would allow fibers to work seamlessly with #1042 , i.e. make async_coropose ops completely inlined, too. This would make the coro compose more effcient for fibers than async_compose, since we skip a bunch of moves & copies.

asio/include/asio/fiber.hpp
```diff
@@ -0,0 +1,1553 @@
...
+#define ASIO_FIBER_HPP
+
+#if defined(_MSC_VER) && (_MSC_VER >= 1200)
+# pragma once
+#endif // defined(_MSC_VER) && (_MSC_VER >= 1200)
+
+#include "asio/detail/config.hpp"
+
+#include <boost/context/fiber.hpp>
+#include <optional>
```
vinniefalco, line 15: Does this require C++17?

asio/include/asio/fiber.hpp
```diff
@@ -0,0 +1,1553 @@
...
+
+namespace asio
+{
+
+namespace detail
+{
+
+struct fiber_handle
+{
+  bool did_suspend{false};
```
vinniefalco, line 39: Maybe put this bool last so it can be packed

asio/include/asio/fiber.hpp
```diff
@@ -0,0 +1,1553 @@
...
+    {
+      Executor executor;
+      Func func;
+      CompletionHandler handler;
+
+      fiber_runner(cancellation_slot slot, Executor executor, Func && func, CompletionHandler && handler)
+          : fiber_handle(slot), executor(std::move(executor)),
+            func(std::forward<Func>(func)), handler(std::forward<CompletionHandler>(handler)) {}
+
+      auto complete()
```
vinniefalco, line 155: `auto` return is C++17?

asio/include/asio/fiber.hpp
```diff
@@ -0,0 +1,1553 @@
...
+                [this, p = std::move(p)]() mutable
+                {
+                  lifetime = std::move(p);
+                  fiber = std::move(fiber).resume();
+                });
+
+      }
+    };
+
+    auto runner = std::make_unique<fiber_runner>(get_associated_cancellation_slot(completion_handler),
```
vinniefalco, line 193: `make_unique` is C++14?

asio/include/asio/fiber.hpp
klemens-morgenstern: It does atm, I guess it should work with C++11 (which is what boost.context requires)

asio/include/asio/fiber.hpp
klemens-morgenstern: See 8190b17, I went down to C++14. `experimental::deferred` requires C++14 support.

asio/include/asio/fiber.hpp
vinniefalco: I don't like the look of this one single bit:
```
executor(std::move(executor))
```

kashirin-alex: It most likely be under a problem with -Wshadow (includes some other ARGS)

vinipsmaker:
Boost.Fiber concurrency is a poor match for IO operations. You should fix Boost.Fiber before proposing integration with Boost.Asio. The only vocabulary lacking in Boost.Fiber for appropriate use in IO contexts is vocabulary to deal with cancellation of operations, so the easy should be that hard. Once the fix is in place, then I think it's worth to discuss integration with Boost.Asio.

vinipsmaker:
Okay, the name misled me. I see now that this PR has no relation to Boost.Fiber.

Now for a more serious review, I don't se a fiber here. Fibers mean the fiber concurrency model and I see zero occurrences of any of that here here (e.g. join, mutexes, condition variables, ...).

Boost.Asio has enough vocabulary to implement fiber concurrency. I have done so at least twice already:

* C++: https://github.com/blinktrade/iofiber
* Lua (Lua bindings for Boost.Asio): https://emilua.gitlab.io/docs/api/0.4/tutorial/fiber.html

You should just stick to the name coroutine (or better: stackful coroutines) if you don't have any plans to integrate real fibers in this work.

vinniefalco:
Vincius my buddy my pal! How's trix? Will you perhaps write a review for Boost.Mustache? Miss you on the list!

vinipsmaker:
> Vinicius my buddy my pal!

Hi, Vinnie

> How's trix?

I don't remember any trix. What's that?

> Will you perhaps write a review for Boost.Mustache? Miss you on the list!

I'll be busy. Can't make it. Sorry.

vinniefalco:
> I don't remember any trix. What's that?

https://en.wiktionary.org/wiki/how%27s_tricks

PR #1046, opened 2022-04-25
Explicitly state `shutdown` thread-safety

MiguelCompany:
We are having a nice discussion [here](https://github.com/eProsima/Fast-DDS/pull/2646#issuecomment-1104914775) regarding the thread-safety of `shutdown(what, ec)` with respect to `receive_from`. Looking at the code, I see no reason for the `shutdown` socket operation to be thread-unsafe, so this PR adds the `shutdown` to the list of thread-safe ones.


MiguelCompany:
It seems this was included in 566bda27670dca6bbaaa217dfeebc6182b00309c, so closing this PR.

PR #1048, opened 2022-04-27
Removed all & race from promise - parallel_group covers most of it now.

klemens-morgenstern:
Removed the race / all functions, as they're covered by parallel_group. Should be global functions if anything.

Should now be C++17.


chriskohlhoff:
Likely addressed by commit 6d48fccb2, released in asio 1.23.0.

PR #1058, opened 2022-05-16
Fix there is no warning number 'XXXX' in windows

exief:
In newer versions of MSVC a number of warning numbers which are disabled in this header are removed so builds using boost asio gets spammed with "there is no warning number 'XXXX'"

Current production builds at work contain this warning in excess of 50,000 times in our build output


chriskohlhoff:
Likely addressed by commit 7fd8033c2, released in asio 1.23.0.

PR #1065, opened 2022-05-30
[windows/posix] Converting move

klemens-morgenstern:
Added converting move to posix & windows classes. I need that for process v2.


chriskohlhoff:
Likely addressed by commit 8a6a6f3ec, released in asio 1.23.0.

PR #1083, opened 2022-07-09
Added error_as_deferred.

klemens-morgenstern:
Heya Chris,

I had an idea: make `error_code` & `exception_ptr` completion tokens, so that the following is legal:

```cpp
asio::steady_timer tim{co_await asio::this_coro::executor};
asio::error_code ec;

co_await tim.async_wait(ec);

```

This would make this much more intuitive because it looks like the sync APIs.

### DIFF
--- a/asio/include/asio/error_as_deferred.hpp
+++ b/asio/include/asio/error_as_deferred.hpp
@@ -1,0 +1,147 @@
+//
+// error_as_deferred.hpp
+// ~~~~~~~~~~~~
+//
+// Copyright (c) 2003-2022 Christopher M. Kohlhoff (chris at kohlhoff dot com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+
+#ifndef ASIO_ERROR_DEFERRED_HPP
+#define ASIO_ERROR_DEFERRED_HPP
+
+#if defined(_MSC_VER) && (_MSC_VER >= 1200)
+# pragma once
+#endif // defined(_MSC_VER) && (_MSC_VER >= 1200)
+
+
+#include "asio/deferred.hpp"
+#include "asio/error.hpp"
+
+#include "asio/detail/config.hpp"
+
+#if (defined(ASIO_HAS_STD_TUPLE) \
+    && defined(ASIO_HAS_DECLTYPE) \
+    && defined(ASIO_HAS_VARIADIC_TEMPLATES)) \
+  || defined(GENERATING_DOCUMENTATION)
+
+#include <tuple>
+#include "asio/associator.hpp"
+#include "asio/async_result.hpp"
+#include "asio/detail/type_traits.hpp"
+#include "asio/detail/utility.hpp"
+
+#include "asio/detail/push_options.hpp"
+
+namespace asio {
+
+
+template <typename Signature>
+class async_result<asio::error_code, Signature>
+{
+public:
+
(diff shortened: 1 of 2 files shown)


asio/src/tests/unit/error_as_deferred.cpp
```diff
@@ -0,0 +1,70 @@
...
+  timer.expires_after(asio::chrono::seconds(0));
+  timer.async_wait(ec)([&count]{count ++;});
+  ASIO_CHECK(count == 0);
+  ASIO_CHECK(!ec);
+
+  ioc.run();
+
+  ASIO_CHECK(!ec);
+  ASIO_CHECK(count == 1);
+#endif // defined(ASIO_HAS_STD_TUPLE)
```
sehe, line 40: this comment seems out of sync

asio/src/tests/unit/error_as_deferred.cpp
```diff
@@ -0,0 +1,70 @@
...
+  timer.async_wait(ec)([&count]{count ++;});
+  timer.cancel();
+  ASIO_CHECK(count == 0);
+  ASIO_CHECK(!ec);
+
+  ioc.run();
+
+  ASIO_CHECK(ec == asio::error::operation_aborted);
+  ASIO_CHECK(count == 1);
+#endif // defined(ASIO_HAS_STD_TUPLE)
```
sehe, line 61: ditto

asio/src/tests/unit/error_as_deferred.cpp
```diff
@@ -0,0 +1,70 @@
...
+}
+
+void error_as_deferred_test_error()
+{
+#if defined(ASIO_HAS_VARIADIC_TEMPLATES)
+  asio::io_context ioc;
+  asio::system_timer timer(ioc);
+  asio::error_code ec;
+  int count = 0;
+  timer.expires_after(asio::chrono::seconds(1000));
```
sehe, line 51: Shower thought: do we need the large timeout? I have an inkling that since `ioc.run()` happens after the _sync_ cancel() it should always be canceled in time.

asio/include/asio/error_as_deferred.hpp
```diff
@@ -0,0 +1,147 @@
...
+  struct deferred_intercept_error_code
+  {
+    asio::error_code &ec;
+
+    template<typename ... Args>
+    auto operator()(asio::error_code ec_, Args && ... args) const
+      -> deferred_values<Args...>
+    {
+      ec = ec_;
+      return asio::deferred_t::values(std::forward<Args>(args)...);
```
sehe, line 54: Would there be performance overhead in using this completion mechanism over, say, directly using `redirect_error(use_awaitable,ec)`?

sehe (commented):
Very clever idea. I like it.

As always, I dread that it might be _too clever_. I don't think there could be ambiguous overloads, though. That leaves potential performance ramifications?

Even if overhead is non-zero, the convenience of this pattern seems worth it.

mzimbres:
I think the usefulness of this feature is more evident in the context of awaitable operators, for example, this is a code snipet in my project

```
      boost::system::error_code ec1, ec2;
      co_await (
         conn->async_run(ep, {}, net::redirect_error(net::use_awaitable, ec1)) &&
         conn->async_exec(req, adapt(), net::redirect_error(net::use_awaitable, ec2))
      );
```
which is annoyingly verbose. I would have prefered to write

```
      boost::system::error_code ec1, ec2;
      co_await (conn->async_run(ep, {}, ec1) && conn->async_exec(req, adapt(), ec2));
```

Issue #1098, opened 2022-07-27
ASIO fails on stringview errors when compiling with QNX-compiler and using -std=c++14


ngroendyk:
It has been consistently observed when building using the QNX compiler (when the language is set to the C++14 standard) that the compiler-set #define "__cplusplus" is set to 201402

This causes the asio header library to include C++17 features that are not yet enabled/configured by the toolchain (since it is configured for C++14), resulting in stringview failures.

chriskohlhoff:
The proposed changes seem to confuse valid `__cplusplus` values with the values used by C++ feature macros (like `__cpp_lib_experimental_string_view`).

In any case, i would expect that this change would be a pure addition (e.g. using `#if defined(__QNXTO__) ...`) and not alter the behaviour for other compilers.

In the meantime, you may change your build settings to define `ASIO_DISABLE_STD_STRING_VIEW` and `ASIO_DISABLE_STD_EXPERIMENTAL_STRING_VIEW as needed.

PR #1099, opened 2022-07-27
Prevent C++17 stringview inclusion in C14 build-configs

ngroendyk:
BUGFIX_ISSUE_1098

* This is a fix for: https://github.com/chriskohlhoff/asio/issues/1098
* According to the C++17 lib-extensions document found at:
  https://en.cppreference.com/w/cpp/experimental/lib_extensions
  StringView was not included until compiler __cplusplus version 201411.
  Examining the source reveals that this compiler condition is evaluated at
  201402. This change adjusts the check from 201402 to 201411.


chriskohlhoff:
See #1098.

PR #1105, opened 2022-08-07
verify all standalone headers

ClausKlein:
A `CMake` project may be used to verify and install a standalone `asio` interface library.
The `CMake export package` is installed to and may be imported from other `CMake` projects.


ClausKlein:
Old code base not longer supported.

see https://github.com/ClausKlein/asio/pull/4

PR #1114, opened 2022-08-28
Fixed redirect_error for deferred.

klemens-morgenstern:
I am trying to make redirect_error work for tokens that don't have a `return_type` defined, such as `deferred`.


chriskohlhoff:
Likely addressed by commit b267e0ceb, released in asio 1.28.0.

PR #1116, opened 2022-08-28
Promise redux

klemens-morgenstern:
Simplified promise & made it conform to the `async_op` concept, so that it can be directly `co_await`ed.


chriskohlhoff:
Likely addressed by commit 00e5b6aaa, released in asio 1.25.0.

PR #1118, opened 2022-08-28
Coro redux

klemens-morgenstern:
Cleaned up allocators, added deferred and regularized `use_coro`.


chriskohlhoff:
Likely addressed by commit 38e19bfec, released in asio 1.25.0.

PR #1128, opened 2022-09-27
Fix broken get_nt_set_info() for overlapped handles

lstipakov:
When calling handle->release(), ASIO uses
kernel-provided NtSetInformationFile() routine
to remove IOCP for the specific handle. Since this routine is not part of SDK, ASIO retrives the address of that function from NTDLL.DLL and caches it
into nt_set_info_ class member. The fact that
address is cached is checked by comparing
it to 0.

The bug is that class member is not initialized
and contains undefined value, so ASIO assumes
that the address is already retrieved and uses
undefined value, which leads to acces violation.

Fix by initializing nt_set_info_ to zero, as it is done for sockets in win_iocp_socket_service_base.ipp.

Signed-off-by: Lev Stipakov <lev@openvpn.net>


chriskohlhoff:
Likely addressed by commit d35861569, released in asio 1.25.0.

PR #1133, opened 2022-10-04
Made cancelled public on a composed_op.

klemens-morgenstern:
This is needed for many composed operatoins, among them boostorg/beast#2512


chriskohlhoff:
Likely addressed by commit eb4080be5, released in asio 1.25.0.

Issue #1138, opened 2022-10-06
asio config does not detect coroutine support with clang-cl as compiler


Febbe:
When compiling ASIO via clang-cl version 15.0 on Windows, the ASIO config sets `ASIO_HAS_CO_AWAIT` to zero, even if clang actually has it. 

The reason is in [line 2024](https://github.com/boostorg/asio/blob/0af7858e7b5603a2415a30b69e16c7ef1d47a5e9/include/boost/asio/detail/config.hpp#L2045-L2070): `if (_MSC_VER >= 1928) && (_MSVC_LANG >= 201705) && !defined(__clang__)`

My proposed solution is to use the `<version>` header. And check for `#if defined(__cpp_lib_coroutine) && __cpp_lib_coroutine == 201902L`

klemens-morgenstern:
It works if you use libc++ by setting `-stdlib=libc++`.

Febbe:
> It works if you use libc++ by setting `-stdlib=libc++`.

Thank you, but not every one can change the STL :/. This might work in some cases for clangd users. But not for the hole project itself, for example.

PR #1145, opened 2022-10-18
Fixed wrong configuration for ASIO_HAS_CO_AWAIT

Febbe:
This adds a more general approach, to check for c++20 coroutines, which takes the feature test macros into account, defined since c++20:    Proposal: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0912r5.html    cppreference: https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros#C.2B.2B20  

I've kept the old macros, in the case compilers + stl-impl add the support, even, when they did not yet set the feature testing macros.
I've also made them more robust: The `#elif` were not mutually exclusive, so I've changed them to `if !defined(ASIO_HAS_CO_AWAIT) && ...`

### DIFF
--- a/asio/include/asio/detail/config.hpp
+++ b/asio/include/asio/detail/config.hpp
@@ -2070,23 +2070,29 @@
 #endif // !defined(ASIO_UNUSED_VARIABLE)
 
 // Support the co_await keyword on compilers known to allow it.
+// Support the co_await keyword on compilers known to allow it.
 #if !defined(ASIO_HAS_CO_AWAIT)
 # if !defined(ASIO_DISABLE_CO_AWAIT)
-#  if defined(ASIO_MSVC)
-#   if (_MSC_VER >= 1928) && (_MSVC_LANG >= 201705) && !defined(__clang__)
+#  if has_include(<version>) && has_include(<coroutine>) && defined(__cpp_coroutines) && (__cpp_coroutines >= 201707L) 
+#   include <version>
+#   if defined(__cpp_lib_coroutine) && (__cpp_lib_coroutine >= 201902L)
+#    define ASIO_HAS_CO_AWAIT 1
+#   endif // defined(__cpp_lib_coroutine) && (__cpp_lib_coroutine >= 201902L)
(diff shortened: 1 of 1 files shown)


Febbe:
Fixes #1138

Febbe:
Waiting for feedback, pipeline and approval. I will rebase / squash (fixup) it then.

madhur4127:
[Interesting test case](https://compiler-explorer.com/z/cx5rWKxqh)

Febbe:
> [Interesting test case](https://compiler-explorer.com/z/cx5rWKxqh)

Can you elaborate? I think to clang13 times, there was no experimental/coroutine include.

madhur4127:
The problem is that using clang13 with libstdc++ will fail in a weird way.

HAS_CO_AWAIT will be true but HAS_COROUTINE is false so if you look at any header say awaitable.hpp it'll fail to find the experimental/coroutine.

Infact the class shouldn't exist because the feature is off.

madhur4127:
I guess what I am trying to hint is that you should also change `HAS_STD_COROUTINE` to pick the right header.

Febbe:
> I guess what I am trying to hint is that you should also change `HAS_STD_COROUTINE` to pick the right header.

This would require another variable like `ASIO_COROUTINE_HEADER`, which is then set to either the plane one or the one in experimental.

madhur4127:
I'm afraid to say this but not every firm can use bleeding edge compilers.

Some firms have a rigorous process to update the compiler and they might be stuck with clang 13.

in this case we can simply check for has_co_await and then if any of coroutine or experimental/coroutine exists we should pick the right one.

This will handle case when using clang13 with exp and non exp coro header

Febbe:
> I'm afraid to say this but not every firm can use bleeding edge compilers.

Doesn't the same apply to bleeding edge / incomplete compiler features? With that logic, no dev must be allowed to use unfinished language features of the current compiler.

I can take a look into this, but if @chriskohlhoff doesn't like it at all, it does not make sense to put more effort into this patch.

chriskohlhoff:
Likely addressed by commit 1511c0e19, released in asio 1.38.2.

PR #1162, opened 2022-11-22
Fix two compilation issues when ASIO_DISABLE_SMALL_BLOCK_RECYCLING is enabled.

sconrad450:
I ran into these two compilation errors, which can be triggered when ASIO_DISABLE_SMALL_BLOCK_RECYCLING is enabled.


chriskohlhoff:
Likely addressed by commit d3fc816e9, released in asio 1.26.0.

PR #1176, opened 2022-12-13
Fixed a potential null pointer dereference

Schoppenglas:
PR fixes a potential null pointer dereference with GCC 12.2 [-Wnull-dereference] -- same condition as in void scheduler::capture_current_exception() or other functions here in scheduler.ipp


avsej:
@chriskohlhoff is there any chance to accept this patch or fix the release build with GCC 12 in any other way?

Schoppenglas:
Resolved.

PR #1191, opened 2023-01-09
windows/basic_console rough implementation.

klemens-morgenstern:
Hi Chris,

I got a very rough implementation for async IO on a console for windows. This would in the next step allow me to make a `open_stdin` function.

I can also add this to process, if you think it's not a fit for asio.


vinipsmaker:
You might be interested in this: https://github.com/ghc/ghc/blob/ghc-8.2.1-release/libraries/base/cbits/inputReady.c#L114

I vaguely remember there were far too many quirks to deal with Windows' standard handles. For instance, whether you can cancel a read operation would depend on the type of the handle (and whether it was instantiated as IOCP... which cannot be modified later on). The only reliable way to cancel operations that I found out at the time I was dealing with Windows console was to create another thread to perform blocking IO and call `CancelSynchronousIo()`. In case you're interested: https://gitlab.com/emilua/emilua/-/blob/a166738c9f181ccce5dc6aaa92e0d2b14d141140/src/system.cpp#L181

Unfortunately I didn't save my notes so I don't have insights to share with you as all my research is now lost.

PR #1211, opened 2023-01-30
Make changes to fix the port binding problem on Mac

adah1972:
- Try and catch exceptions in echo server listener so that the examples are able to report port binding error.
- Change the use of port 55555 to 55557


adah1972:
@chriskohlhoff Would you please take a look? This change is quite important for users on macOS.

PR #1215, opened 2023-01-31
Fix initialization of sqe->user_data for reuse by io_uring_service.

jevolk:
The value of user_data in a submission queue entry is only initialized with with the ring itself, remaining unchanged after use and even through calls to io_uring_get_sqe(3). Many submission paths apply io_uring_sqe_set_data(3) but the remainder are submitted with stale values that can apply completion operations on the wrong queues with the wrong results, causing obscure bugs.


vinipsmaker:
Landed in `master`: https://github.com/chriskohlhoff/asio/commit/0080ec1c9b0991c0bd9356ee73d0401ca8d315dc

Please close the PR.

chriskohlhoff:
Likely addressed by commit 0080ec1c9, released in asio 1.27.0.

PR #1224, opened 2023-02-08
Set SA_RESTART when establishing an UNIX signal handler by default

### LINKED ISSUE #646: Signal handler API has no way to set SA_RESTART
*@nomis commented on Oct 29, 2018, 8:14 PM UTC:*

I'm handling SIGCHLD using asio while also doing writes to standard output and reads/writes to files.

If I receive SIGCHLD during a `read()` or `write()` then it gets interrupted. The simplest way to deal with this is to set `SA_RESTART` when setting up the signal handler, but there is no support for doing this and it doesn't happen by default.

*This issue was moved by [chriskohlhoff](https://github.com/chriskohlhoff) from [boostorg/asio#157](https://github.com/boostorg/asio/issues/157).*

vinipsmaker:
The user doesn't have control over 3rd party library code that he uses in his project. If any code forgets to make use of a macro such as TEMP_FAILURE_RETRY() in *any* syscall, we have a bug.

Signals that interrupt syscalls are a bad default and BSD acknowledged that problem when it changed the default signal semantics to SA_RESTART way back at the release of 4.2BSD. POSIX upstreamed this option through the use of SA_RESTART, and ASIO should be making use of this flag by default (hence this commit).

If any code wants the old behavior, it can opt-out of the new default by defining the macro ASIO_NO_SET_SA_RESTART.

SA_RESTART is only meaningful when we establish a signal handler (asio_signal_handler), so we skip SA_RESTART altogether when sa_handler is set to any other value (SIG_DFL).

Fixes #646

### DIFF
--- a/asio/include/asio/detail/impl/signal_set_service.ipp
+++ b/asio/include/asio/detail/impl/signal_set_service.ipp
@@ -342,6 +342,9 @@ asio::error_code signal_set_service::add(
       struct sigaction sa;
       memset(&sa, 0, sizeof(sa));
       sa.sa_handler = asio_signal_handler;
+#  if !defined(ASIO_NO_SET_SA_RESTART)
+      sa.sa_flags = SA_RESTART;
+#  endif
       sigfillset(&sa.sa_mask);
       if (::sigaction(signal_number, &sa, 0) == -1)
 # else // defined(ASIO_HAS_SIGACTION)
--- a/asio/src/doc/using.qbk
+++ b/asio/src/doc/using.qbk
@@ -400,6 +400,13 @@ functionality, and behaviour of Asio.
       types are not passed across shared library API boundaries.
     ]
   ]
+  [
+    [`ASIO_NO_SET_SA_RESTART`]
+    [
+      Disables the use of the flag `SA_RESTART` when establishing signal
+      handlers.
+    ]
+  ]
 ]
 
 [include platform_macros.qbk]


vinipsmaker:
The real problem with signals is that they're too intrusive. Once you add custom signal handling to your code, every piece of code that was there before will now have to be updated to match the new mindset (any syscall can now be interrupted by `EINTR`). They require you to constantly and explicitly think about signals at every part of the application life cycle. So obviously, most of the code will do the wrong thing in the presence of signals. `SA_RESTART` fixes that (now you just need to think about signals on code that actually deals with signals).

There are places where's not even trivial (trivial meaning wrapping the syscall with a macro such as `TEMP_FAILURE_RETRY()` that will just emulate `SA_RESTART` which is the thing that should be on by default anyway) at all how to manually handle `EINTR`: <http://ewontfix.com/4/>. `SA_RESTART` also fixes this problem.

Rust goes as far as to even wrap all syscalls and swallow `EINTR` altogether. It doesn't even offer a way to opt-out of this behavior (and nobody miss `EINTR`): https://github.com/rust-lang/rust/blob/ef934d9b632b8ac00276558824664c104b92b5f0/library/std/src/io/mod.rs#L380

The only reason why I even added a macro to keep the old behavior is just that you feel more comfortable with merging this PR. Usually I worked with code that was under my full control, so I could use Boost.Asio signal handling facilities. That's not the case for my new project (Lua bindings for Boost.Asio). I really need this PR merged unless I give up on Boost.Asio signal handling facilities and write my own signal `asio::signal_set` that will be a complete copy-paste except for `sa_flags|=SA_RESTART` at `sigaction()`.

vinipsmaker:
https://github.com/chriskohlhoff/asio/commit/50ad5e47793521a213eb461da701e8ddd3d5c70d landed in `master` which is an alternative solution

PR #1226, opened 2023-02-09
Windows: map connection_aborted to ERROR_CONNECTION_ABORTED

masoomshaikh:


phprus:
@masoomshaikh 
Please tell me, without this PR, what kind of errors can there be when using asio?

PR #1237, opened 2023-02-17
Fixes result handling for promise.

### LINKED ISSUE #1204: co_await promise(use_awaitable) does not propagate exceptions (returns bogus value)
Complete example. I expected the promise to throw, but it returns a default constructed int instead:
```c++
#include <iostream>
#include <cassert>
#include <boost/asio/use_awaitable.hpp>
#include <boost/asio/detached.hpp>
#include <boost/asio/deferred.hpp>
#include <boost/asio/co_spawn.hpp>
#include <boost/asio/io_context.hpp>
#include <boost/asio/experimental/promise.hpp>
#include <boost/asio/experimental/use_promise.hpp>

using namespace boost::asio;
using namespace boost::asio::experimental;

awaitable<int> f()
{
    throw std::runtime_error("foo");
    co_return 42;
}

int main()
{
    io_context ioc;
    co_spawn(ioc, [&]() -> awaitable<void>{
        try
        {
            auto promise = co_spawn(ioc, f(), use_promise);
            //int r = co_await promise(deferred); // does not compile
            int r = co_await promise(use_awaitable); // does not throw, triggers assert
            std::cout << r << std::endl; // r is 0
            assert(false);
        }
        catch (const std::exception& e)
        {
            std::cout << e.what() << std::endl;
        }
    }, detached);
    ioc.run();
}
```

klemens-morgenstern:
I've had a bug where the promise used moved-from values when delivering results & the promise is already waiting. This should fix it.

Closes #1204

### DIFF
--- a/asio/include/asio/experimental/impl/promise.hpp
+++ b/asio/include/asio/experimental/impl/promise.hpp
@@ -228,7 +228,12 @@ struct promise_handler<void(Ts...), Executor, Allocator>
     impl_->done = true;
 
     if (impl_->completion)
-      impl_->complete(std::move(ts)...);
+      impl_->apply(
+        [this](std::remove_reference_t<Ts>... ts_)
+        {
+          impl_->complete(std::move(ts_)...);
+        }
+      );
   }
 };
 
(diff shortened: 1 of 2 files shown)


chriskohlhoff:
Likely addressed by commit b1f29a7f7, released in asio 1.28.0.

Issue #1241, opened 2023-02-22
asio::experimental::channel: try_send immediately after cancel always fails


erinacio:
MRE:

```cpp
#include <asio.hpp>
#include <asio/experimental/channel.hpp>
#include <iostream>

asio::awaitable<void> async_main() {
  auto ex = co_await asio::this_coro::executor;
  asio::experimental::channel<void(std::error_code, int)> channel_{ex, 16};
  asio::co_spawn(
      ex,
      [](auto &ch) -> asio::awaitable<void> {
        co_await ch.async_receive(asio::as_tuple(asio::use_awaitable));
      }(channel_),
      asio::detached);
  asio::steady_timer timer{ex, std::chrono::milliseconds{1}};
  // make sure async_receive is awaiting
  co_await timer.async_wait(asio::as_tuple(asio::use_awaitable));
  channel_.cancel();
  bool ok = channel_.try_send(std::error_code{}, 1);
  std::clog << "try send ok? " << ok << std::endl;
}

int main() {
  asio::io_context io{1};
  asio::co_spawn(io, async_main(), asio::detached);
  io.run();
  return 0;
}
```

got

```
try send ok? 0
```

on [asio 1.24.0](https://github.com/chriskohlhoff/asio/tree/asio-1-24-0) and [asio latest master commit](https://github.com/chriskohlhoff/asio/tree/a71f5232d207b4f3bbd253eb1041e30b5e4ea606) with gcc 12.2.1 and clang 15.0.7 on Arch Linux.

 According to the [documentation](https://think-async.com/Asio/asio-1.24.0/doc/asio/reference/experimental__basic_channel/try_send.html), this behavior is not expected as the channel is apparently not full.

erinacio:
#1242 merged externally

PR #1242, opened 2023-02-22
Fix channel try_send failure after cancel

erinacio:
The only case when `impl.send_state_` becomes `waiter` seems to be that an `async_receive` is awaiting on an empty channel, as found in [channel_service.hpp:557](https://github.com/chriskohlhoff/asio/blob/a71f5232d207b4f3bbd253eb1041e30b5e4ea606/asio/include/asio/experimental/detail/impl/channel_service.hpp#L557). As an empty channel, when no one is awaiting on it, whether `send_state_` is `buffer` or `block` should solely depend on whether the channel itself is buffered or not.

Fixes https://github.com/chriskohlhoff/asio/issues/1241


vinipsmaker:
Landed in `master`: https://github.com/chriskohlhoff/asio/commit/efec5c197d29b329e16f98000695a1f98cfa65df

Please close the PR.

PR #1254, opened 2023-03-05
Added public getter for `concurrency_hint` on the io_context.

klemens-morgenstern:
This can be very useful for other libraries, as they can use it to base their own concurrency policy on it.

I specifically needs this for my async_mutex library. That is, my mutex will be much simpler for single-threaded environments, which I would like to check by querying the io_context.

### DIFF
--- a/asio/include/asio/impl/io_context.ipp
+++ b/asio/include/asio/impl/io_context.ipp
@@ -169,6 +169,11 @@ void io_context::service::fork_service(io_context::fork_event)
 }
 #endif // !defined(ASIO_NO_DEPRECATED)
 
+int io_context::get_concurrency_hint() const
+{
+  return impl_.concurrency_hint();
+}
+
 } // namespace asio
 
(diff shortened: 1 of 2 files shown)


vinipsmaker:
> This can be very useful for other libraries

I can confirm that. In the Lua bindings I wrote for Boost.Asio, I've created a service just to attach the concurrency hint to the `io_context`:

https://gitlab.com/emilua/emilua/-/blob/a166738c9f181ccce5dc6aaa92e0d2b14d141140/src/main.ypp#L375

> as they can use it to base their own concurrency policy on it

For my own use, I probe the concurrency hint for the current `io_context` to check whether I can spawn a new thread to call `run()`:

https://gitlab.com/emilua/emilua/-/blob/a166738c9f181ccce5dc6aaa92e0d2b14d141140/src/actor.ypp#L1855

chriskohlhoff:
Likely addressed by commit 79163e4c3, released in asio 1.33.0.

PR #1258, opened 2023-03-07
Add more signal_set_base::flags_t constants

vinipsmaker:
This commit only adds useful POSIX constants that don't conflict with asio's signal handling strategy (e.g. SA_SIGINFO would break asio interface, so it's avoided). Non-POSIX constants specific to certain operating systems (e.g. Linux's SA_UNSUPPORTED) are ignored.


vinipsmaker:
Alternatively, you could use `no_child_stop` instead of `nocldstop` and `no_child_wait` instead of `nocldwait`, but I can't know what style you prefer, so you'll have to choose yourself.

vinipsmaker:
Landed in `master`: https://github.com/chriskohlhoff/asio/commit/11c6cafd3d4397fedb37003f119506a296813044

PR #1259, opened 2023-03-10
only test the windows version when targeting old windows versions

maxbachmann:
On non desktop windows targets `VerSetConditionMask` and `VerifyVersionInfo` are not always available. When targeting platforms >= Windows Vista we can skip this check, since it is guaranteed to be true.


maxbachmann:
ping @chriskohlhoff : any chance of getting this merged?
We need this patch for xbox support.

maxbachmann:
apparently this was merged without commenting here: https://github.com/chriskohlhoff/asio/commit/352177ef98fb9d10513a05d0e139c3f7cb918b57

PR #1267, opened 2023-03-25
fix call buffer(const const_buffers_1&,size_t)  

marian556:
and call buffer(const mutable_buffers_1&,size_t)  to get it working 
like they did in boost 1.79/asio 1.22 and before (compatibility issue) 
unless ASIO_BUFFER_ARG_BUFFERS_1_IS_BUFFERSEQUENCE is defined. 
There is a breaking undisclosed runtime change introduced in Asio 1.24.0 / Boost 1.80 : 
"Added buffer() overloads for contiguous containers, such as std::span"
That change invokes incorrectly the generic overloads of function buffer 
designed for Generic Contiguous  Sequences/Containers that 
const_buffers_1 meets technically ,but the sequence creates buffer of buffers 
pointing to const_buffers_1 class location , rather than to its underlaying data

### DIFF
--- a/asio/include/asio/buffer.hpp
+++ b/asio/include/asio/buffer.hpp
@@ -934,6 +934,26 @@ ASIO_NODISCARD inline ASIO_MUTABLE_BUFFER buffer(
         ));
 }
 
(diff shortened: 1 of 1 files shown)


chriskohlhoff:
Likely addressed by commit fb451d46d, released in asio 1.28.0.

PR #1271, opened 2023-03-27
Fix `ASIO_ASSUME` definition on older compilers.

### LINKED ISSUE #1270: `ASIO_ASSUME` definition is broken on older compilers
https://github.com/chriskohlhoff/asio/blob/f693a3eb7fe72a5f19b975289afc4f437d373d9c/asio/include/asio/detail/config.hpp#L2146-L2157

1. gcc 4.4 does not support __builtin_unreachable, which makes the `__GNUC__` branch definition to fail.

2. clang 3.5 doesn't support `__builtin_assume` and uses the last fallback option. There, `ASIO_ASSUME` is defined without arguments, meaning it expands to incorrect code like `(void)0(base != 0)`. This also affects any other compilers that are not handled by prior branches.

This also affects Boost version, including Boost 1.82 beta1. In particular, it prevents Boost.Log compilation with the compilers mentioned above. Please, fix before the Boost 1.82 release.

Lastique:
1. gcc 4.4 does not support `__builtin_unreachable`, which makes the `__GNUC__` branch definition to fail.

2. clang 3.5 doesn't support `__builtin_assume` and uses the last fallback option. There, `ASIO_ASSUME` is defined without arguments, meaning it expands to incorrect code like `(void)0(base != 0)`. This also affects any other compilers that are not handled by prior branches.

Fixes https://github.com/chriskohlhoff/asio/issues/1270.

Please, also merge this to Boost before Boost.1.82 is released.

### DIFF
--- a/asio/include/asio/detail/config.hpp
+++ b/asio/include/asio/detail/config.hpp
@@ -2149,11 +2149,11 @@
 # if __has_builtin(__builtin_assume)
 #  define ASIO_ASSUME(expr) __builtin_assume(expr)
 # endif // __has_builtin(__builtin_assume)
-#elif defined(__GNUC__)
+#elif defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 5))
 # define ASIO_ASSUME(expr) if (expr) {} else { __builtin_unreachable(); }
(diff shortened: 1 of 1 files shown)


chriskohlhoff:
Likely addressed by commit 451a97eb1, released in asio 1.28.0.

PR #1276, opened 2023-04-06
Add constexpr support to ip::address* and ip::network* classes

rcombs:
This is split into 2 segments: implementation changes that make the code more constexpr-compatible, and changes that actually add constexpr tags. The new functionality is unit-tested extensively. Most functions are only made constexpr when building header-only, since they're either defined in .ipp files or depend on functions that are.


rcombs:
The initial version of this PR ran into some issues on the CI due to Windows using a union type for `in_addr`, where the 32-bit int member doesn't come first. This seems to make it impossible to initialize and assign to in constexpr code pre-C++20. The current version addresses this largely by requiring C++20 for constexpr on address_v4.

This could alternately be addressed by using `bytes_type` as the internal storage format for `address_v4`, and bit-casting/memcpying to/from `in_addr` when calling `inet_pton` and `inet_ntop`. This could also be applied to IPv6; in both cases, it'd allow us to support constexpr on older C++ versions; it should be ABI-compatible with the existing code, though it's possible that some cursed platform out there has auxiliary padding or some other nonsense in the `in[6]_addr` structs.

PR #1280, opened 2023-04-14
Fix using completion handlers with rvalue-ref-qualified call operator with promise

### LINKED ISSUE #1279: Bug: promise cannot be used with completion handlers with an rvalue-ref qualified call operator

mika-fischer:
Fixes #1279

### DIFF
--- a/asio/include/asio/experimental/impl/promise.hpp
+++ b/asio/include/asio/experimental/impl/promise.hpp
@@ -64,10 +64,10 @@ struct promise_impl<void(Ts...), Executor, Allocator>
   Executor executor;
 
   template<typename Func, std::size_t... Idx>
-  void apply_impl(Func f, asio::detail::index_sequence<Idx...>)
+  void apply_impl(Func&& f, asio::detail::index_sequence<Idx...>)
   {
     auto& result_type = *reinterpret_cast<promise_impl::result_type*>(&result);
-    f(std::get<Idx>(std::move(result_type))...);
+    std::forward<Func>(f)(std::get<Idx>(std::move(result_type))...);
   }
 
   using allocator_type = Allocator;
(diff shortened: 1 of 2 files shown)


asio/include/asio/experimental/impl/promise.hpp
```diff
@@ -64,10 +64,10 @@ struct promise_impl<void(Ts...), Executor, Allocator>
   Executor executor;
 
   template<typename Func, std::size_t... Idx>
-  void apply_impl(Func f, asio::detail::index_sequence<Idx...>)
+  void apply_impl(Func&& f, asio::detail::index_sequence<Idx...>)
   {
     auto& result_type = *reinterpret_cast<promise_impl::result_type*>(&result);
-    f(std::get<Idx>(std::move(result_type))...);
+    std::forward<Func>(f)(std::get<Idx>(std::move(result_type))...);
```
kashirin-alex, line 70: This part makes no sense

asio/include/asio/experimental/impl/promise.hpp
```diff
@@ -64,10 +64,10 @@ struct promise_impl<void(Ts...), Executor, Allocator>
   Executor executor;
 
   template<typename Func, std::size_t... Idx>
-  void apply_impl(Func f, asio::detail::index_sequence<Idx...>)
+  void apply_impl(Func&& f, asio::detail::index_sequence<Idx...>)
```
kashirin-alex, line 67: Here a const ref can do the trick - 'func' remains in scope and not a subject for a change.

asio/include/asio/experimental/impl/promise.hpp
mika-fischer: Either that or take by value (i.e. move-construct the function) and move. Everything else does not work with rvalue-ref-qualified call operators on completion handlers. Or am I missing something?

If it's between these two options, I'd avoid the (unnecessary) move-construction.

asio/include/asio/experimental/impl/promise.hpp
mika-fischer: Cannot call `operator(...) &&` on a const ref...

chriskohlhoff:
Likely already fixed / superseded by commit b1f29a7f7, released in asio 1.28.0.

PR #1282, opened 2023-04-18
solaris compile fix

tkoecker:
The use of epoll_reactor in dev_poll_reactor.hpp breaks the build on Solaris 10.
I assume this is a copy paste from epoll_reactor.hpp error. This change fixes the build.


chriskohlhoff:
Likely addressed by commit 267775c43, released in asio 1.28.1.

PR #1283, opened 2023-04-18
Prevent new_registration leak on error paths

gjasny:
Since 50ad5e47793521a213eb461da701e8ddd3d5c70d the `new_registration` is leaked in the error paths. Maybe the allocation of `new_registration` could be moved closer to the usage, instead.


chriskohlhoff:
Likely addressed by commit 7e8baf64f, released in asio 1.28.1.

PR #1288, opened 2023-04-26
Enable interrupts to be work again after poll() has been called.

blackbox-tech:
I would like to merge this change because I have a custom event loop that calls poll() whenever the reactor is interrupted.  However, because poll() does not reset ```task_interrupted_``` after it has run the task, posting events to the io_context from another thread does not interrupt the io_context like calling run() or wait() does.

I hope its an acceptable bug fix, as this change is essentially making the behavior of poll() match the state change that read() and wait() make,  although I accept my use case integrating asio inside an external event-loop is a bit unusual.


blackbox-tech:
Ignore, this fix it does not help as ```task_cleanup``` always sets ```task_interrupted_``` back to ```true``` anyway.

PR #1301, opened 2023-05-15
Remove broken links to echo examples under C++14

slowriot:
The documentation includes links to echo examples under the C++14 section, but these examples don't exist (and appear to have just been duplicated from the C++11 section).

Broken links can be seen in the live documentation: https://www.boost.org/doc/libs/1_82_0/doc/html/boost_asio/examples/cpp14_examples.html / https://www.boost.org/doc/libs/1_82_0/doc/html/boost_asio/example/cpp14/echo/async_tcp_echo_server.cpp etc.

### DIFF
--- a/asio/src/doc/examples.qbk
+++ b/asio/src/doc/examples.qbk
@@ -586,19 +586,6 @@ completion token.
 * [@../src/examples/cpp14/deferred/deferred_7.cpp]
 
 
-[heading Echo]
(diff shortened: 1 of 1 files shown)


chriskohlhoff:
Likely already fixed / superseded by commit 0fd3cf3dd, released in asio 1.20.0.

PR #1313, opened 2023-06-12
Workaround for experimental::coro compile error with MSVC

cbrl:
This addresses a problem with MSVC where it incorrectly reports a syntax error when compiling code that uses `asio::experimental::coro`. There is an [existing bug report](https://developercommunity.visualstudio.com/t/Syntax-Error-when-using-boost::asio::exp/10224448) on the Microsoft developer community. However, this can also be fixed in ASIO itself by renaming a few captured lambda variables.


cbrl:
Fixed in a09dde9d6b561f21db9e7abfedca4cdc210be8e9

PR #1323, opened 2023-06-30
Check for null target in query/require/prefer

### LINKED ISSUE #1322: any_executor().query() segfaults instead of throwing bad_executor
Motivating example (incorrect, but an easy mistake to make):
```c++
asio::any_io_executor ex;
asio::ip::tcp::socket sock(ex);
```
Prior to https://github.com/chriskohlhoff/asio/commit/9d420678b4f89d20585f42f85deb0838e3597991 this would throw `asio::execution::bad_executor`; after that commit it just segfaults.

Online example using Boost.Asio (1.80 vs 1.81): https://godbolt.org/z/WGvMEWr8P

We noticed this because clang-tidy started warning that `object_fns_` could be indirected while null.

Presumably `query`/`require`/`prefer` need to check for null `target_` the same way `execute` does?

ecatmur:
and throw bad_executor if empty. Fixes #1322.

### DIFF
--- a/asio/include/asio/execution/any_executor.hpp
+++ b/asio/include/asio/execution/any_executor.hpp
@@ -1823,9 +1823,17 @@ public:
         >::value
       >::type* = 0) const
   {
-    typedef find_convertible_property<Property> found;
-    prop_fns_[found::index].query(0, object_fns_->target(*this),
-        &static_cast<const typename found::type&>(p));
+    if (target_)
+    {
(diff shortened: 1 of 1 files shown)


ecatmur:
Possibly fixed by 6c4a16680caa5e63e742f7041cf1fccaa30f1841, checking

ecatmur:
Confirming fixed by 6c4a16680caa5e63e742f7041cf1fccaa30f1841.

PR #1324, opened 2023-07-01
MS-DOS compatibility

jwt27:
This small patch will make asio compile for 32-bit DOS (djgpp), with the Watt32
socket library.

Some IPV6 option macros are missing, I don't know exactly how complete the
ipv6 implementation is.  Defining those to 0 gets it to compile, but using them
will of course do nothing.

I tried some of the example programs.  They compiled with no warnings, and seem
to run fine.


jwt27:
I did discover a bug in libc that prevents `ioctl()` arguments from being
passed through to the socket library.  So the workaround is to use
`ioctlsocket()`, but it takes a `char*`, so requires extra casts.

I sent in a patch to the mailing list.  If that gets picked up soon then I
might drop the last commit, since it just adds extra clutter.


This also uncovered a bug/quirk in the way the socket library handles a
non-blocking `connect()`.  Working on a PR for that now.

jwt27:
Better solution: `ioctlsocket()` now takes a `void*` argument, so there's no need for extra casts (ref: https://github.com/gvanem/Watt-32/pull/81).

The issue with non-blocking `connect()` also got fixed (ref: https://github.com/gvanem/Watt-32/pull/80).

Issue #1366, opened 2023-10-02
tss_ptr does not play well with fibers that migrate threads


bgemmill:
When using boost::fiber with fibers with boost::asio, sometimes we get segfaults because the tss usage inside asio is not fiber-tolerant.

For reference, boost::fiber warns of this, saying ["it would be problematic to migrate a fiber that relies on thread-local storage."](https://www.boost.org/doc/libs/1_83_0/libs/fiber/doc/html/fiber/migration.html#ftn.fiber.migration.f1) 

The issue is that boost::asio uses tss_ptr a lot while calling `detail::thread_context::top_of_thread_call_stack`, and this can lead to segfaults.

There are a couple workarounds, but I'm not sure how well these would map to boost::asio.

The first is that one can force the compiler not to cache the address of the thread_local, effectively always getting the *current* thread's tss. Stack overflow has a good discussion of it [here](https://stackoverflow.com/questions/75592038/how-to-disable-clang-expression-elimination-for-thread-local-variable), where the noinline attribute does the heavy lifting, looking something like:
```
      __attribute__((noinline))
      static some_data_type& singleton() {
        static thread_local some_data_type tss;
        return tss;
      }
```

Since boost::asio wants to keep track of the callback stack though, I think we'd either need [fiber-specific storage](https://www.boost.org/doc/libs/1_83_0/libs/fiber/doc/html/fiber/fls.html) to maintain the stack, or some way to disable the stack entirely for fiber users.

thed636:
Bumped in the problem too while using the `awaitable` with the `thread_pool`. Was misled a lot by [the post](https://stackoverflow.com/questions/75183983/do-boostasio-c20-coroutines-support-multithreading) on StackOverflow.

PR #1367, opened 2023-10-03
make tss_ptr safe for fibers that migrate threads

bgemmill:
fix https://github.com/chriskohlhoff/asio/issues/1366


thed636:
Bumped in the problem too while using the `awaitable` with the `thread_pool`. Was misled a lot by [the post](https://stackoverflow.com/questions/75183983/do-boostasio-c20-coroutines-support-multithreading) on StackOverflow.

PR #1387, opened 2023-11-29
ionotify() implementation for QNX based on QNX pulse communcation

VladCimpianu:
The current ASIO socket implementation on QNX uses two different processes for communication. The io_pkt (or in newer version io_sock) and the pipe daemon. Involving additional daemons in basic communication paths is always less than optional with respect to latency and CPU load.
While io_pkt implements the real socket functionality, the pipe daemon is only used to implement the file descriptor select functionality. ASIOs default select_reactor implementation on QNX uses the pipe_select_interrupter, which performs the interactions with the pipe daemon. The proposed patch (initially implemented by Blackberry) implements its own ionotify_reactor, which builds on top of the ionotify() system call and asynchronous QNX pulse communication. As a result, the interaction with the additional pipe daemon is not needed anymore.

### DIFF
--- a/asio/include/asio/detail/config.hpp.orig
+++ b/asio/include/asio/detail/config.hpp.orig
@@ -884,16 +884,6 @@
 # endif // !defined(ASIO_HAS_DEV_POLL)
 #endif // defined(__sun)
 
-// QNX: ionotify.
-#if defined(__QNXNTO__)
-# if !defined(ASIO_HAS_IONOTIFY)
-#  if !defined(ASIO_DISABLE_IONOTIFY)
-#   define ASIO_HAS_IONOTIFY 1
-#   define ASIO_DISABLE_EPOLL
-#  endif // !defined(ASIO_DISABLE_IONOTIFY)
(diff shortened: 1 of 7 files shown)


VladCimpianu:
Hello @chriskohlhoff @klemens-morgenstern @vinipsmaker @MarcelRaad 

Sorry to notify you like this but I cant find any way to contact you. 
Is it possible to review this PR? 

Thank you!

MarcelRaad:
I'm not a maintainer, sorry.

VladCimpianu:
> I'm not a maintainer, sorry.

Do you have any contact from any maintainer?

olegv11:
At NVIDIA we've encountered problems with QNX's select_reactor due to the limit on the max value of file descriptors that select accepts. I've written a custom reactor based on poll() (which I can submit a patch for if there's interest). But ionotify(), a native mechanism for QNX, sounds like a much better solution. It'd be great to see this patch merged.

@VladCimpianu 
About the patch itself. It seems like config.hpp.orig was accidentally added to the commit. There's also an __SRCVERSION macro that isn't used anywhere else in the project.

PR #1406, opened 2024-01-09
Add macro ASIO_DISABLE_FIONBIO

caloni:
On FreeBSD, sockets received from capsicum processes will return ENOTCAPABLE on ioctls. In this scenario, fcntl() is the preferred approach to put the file descriptor in non-blocking mode.

Furthermore, nested kqueues are prevented w/o this patch. FIONBIO is not implemented by kqueue file descriptors. So it's impossible to use ASIO (nested kqueue through posix::stream_descriptor) to await for a procdesc to finish, as an example.

It's also generally unsafe to perform any ioctl on file descriptors received from untrusted processes. Linux is slowly adding support for this type of sandbox through Landlock, so it's not just a FreeBSD problem.

FIONBIO already caused problems to ASIO in this past (cf. commit 607f99ba80df71f1a96f0df71725059ee71e46ff). This patch only exposes a knob that users can also resort to when FIONBIO fail on them.

### DIFF
--- a/asio/include/asio/detail/impl/descriptor_ops.ipp
+++ b/asio/include/asio/detail/impl/descriptor_ops.ipp
@@ -63,11 +63,11 @@ int close(int d, state_type& state, asio::error_code& ec)
       // current OS where this behaviour is seen, Windows, says that the socket
       // remains open. Therefore we'll put the descriptor back into blocking
       // mode and have another attempt at closing it.
-#if defined(__SYMBIAN32__) || defined(__EMSCRIPTEN__)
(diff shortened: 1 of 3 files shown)


VladCimpianu:
Hello!

Sorry to bother like this, but somehow do you have any contact from any responsible of Reviews and Merges into this repo? like: @chriskohlhoff @klemens-morgenstern @vinipsmaker @MarcelRaad
I tried to find any way to contact them but can't find anything.
I made a PR 3 months ago and still didn't get any review.

Thank you!

PR #1436, opened 2024-02-28
Use 2.2 Specification when initializing Winsock

### LINKED ISSUE #930: Initialize winsock with version 2.2
`asio::detail::winsock_init` initalizes winsock with v2.0.

Not sure whether this makes any difference for asio and the world in general, I just think it's a good idea to default to the latest version.

* v2.0 was only an internal draft version.
* v2.1 was the first public winsock 2 specification.
* Nobody needs 16-bit OS support these days anymore?
* v2.2 itself is ages old, and supported back to Windows 95 OSR2.

xavier2k6:
Closes #930.

### DIFF
--- a/asio/include/asio/detail/winsock_init.hpp
+++ b/asio/include/asio/detail/winsock_init.hpp
@@ -47,7 +47,7 @@ protected:
   ASIO_DECL static void throw_on_error(data& d);
 };
 
-template <int Major = 2, int Minor = 0>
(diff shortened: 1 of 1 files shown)


xavier2k6:
Changes were implemented in https://github.com/chriskohlhoff/asio/commit/ee3bc5c89d310ee994073ba68ca35cb64332f8fd & will be included in Boost 1.85.0

PR #1445, opened 2024-03-13
Use WSAConnect when socket has been created by WSASocket

avsej:
Otherwise it is possible that the ::connect() might return code

    10038 An operation was attempted on something that is not a socket.

In particular it happens when trying to use ASIO inside Ruby native extension on Windows (which is using MinGW).


avsej:
I will try to dig deeper and rather find the cause in this particular integration, because standalone test compiles with ASIO just find. It should be something else, maybe Ruby intercepts some winsock calls or something like that.

PR #1447, opened 2024-03-18
asio.hpp: Added IWYU Pragma export

FlorianReimold:
`asio.hpp` is an umbrella header, however it needs to be marked as such when using static IWYU (Include What You Use) checks like recent `clang-tidy` versions. Without those `IWYU pragma`s, clang tidy will complain a lot:

image

Clang tidy include cleaner documentation can be found here: https://clangd.llvm.org/design/include-cleaner


Jezza2:
Any progress on this? I'd like it too!

Issue #106, opened 2016-03-12
socket option for REUSEPORT?


rpavlik:
Or, this might be combined with reuse_address, though I'm not sure the structure allows for that very well - but apparently, in at least newer Linux kernels, this is "a thing" and for the traditionally-expected reuse address behavior you have to do both, at least according to this: http://stackoverflow.com/a/25193462/265522

PR #1455, opened 2024-03-28
Set SO_REUSEPORT on Linux if available

stephanlachnit:
Closes https://github.com/chriskohlhoff/asio/issues/106.

Since Linux 3.9 the SO_REUSEPORT option is also available. This allows for e.g. to use several UDP sockets listening to the same broadcasts within a process.

I'm not sure if it would be better to add a `asio::socket_base::reuse_port` instead, since one might want to set one but not the other.


stephanlachnit:
Reading through https://stackoverflow.com/a/14388707, this actually isn't correct, `SO_REUSEPORT` works different for UDP on linux, as in it does load balancing.

stephanlachnit:
Actually reading through it again, I am not sure anymore. The load balancing part of `SO_REUSEPORT` is actually nice, but there is one restriction namely that ports can be hijacked anymore (see https://lwn.net/Articles/542629/).

PR #1457, opened 2024-04-05
Install .pc file to /usr/share instead of /usr/lib

DarthGandalf:
because there are no .so files


chriskohlhoff:
Likely addressed by commit a6da88e54, released in asio 1.32.0.

DarthGandalf:
Well, yes, because it is this same commit, by me :)

Probably you had merged this PR manually without closing this PR

PR #1458, opened 2024-04-06
using constexpr inline for global tag types.

klemens-morgenstern:
Hi Chris,

I am working on experimental module support for boost and it complains about exporting none-inline global variables. 

Declaring them inline seems like the correct thing to do anyhow.


chriskohlhoff:
Likely addressed by commit efc2f7754, released in asio 1.38.2.

PR #1461, opened 2024-04-10
 Fix Apple clang build warning/error on iOS/tvOS

biodranik:
```
asio/ip/impl/address_v6.ipp:284:28: error: implicit conversion loses integer precision: 'unsigned long' to 'scope_id_type' (aka 'unsigned int') [-Werror,-Wshorten-64-to-32]
```


chriskohlhoff:
Likely addressed by commit 447ae53dc, released in asio 1.32.0.

PR #1487, opened 2024-06-05
Fix typo in Cancellation documentation

Afforix:
It seems to me that the `total` guarantee should satisfy the `terminal` guarantee, not the `total` guarantee again.


chriskohlhoff:
Likely addressed by commit 84e89e5a4, released in asio 1.32.0.

PR #1489, opened 2024-06-09
Fix asio::ip::tcp::resolver::query assignment

georgthegreat:
Fixes:

```
error: object of type 'asio::ip::tcp::resolver::query' (aka 'basic_resolver_query<asio::ip::tcp>') 
cannot be assigned because its copy assignment operator is implicitly deleted
```

### DIFF
--- a/asio/include/asio/ip/basic_resolver_query.hpp
+++ b/asio/include/asio/ip/basic_resolver_query.hpp
@@ -212,21 +212,10 @@ public:
     hints_.ai_next = 0;
   }
 
-  /// Copy construct a @c basic_resolver_query from another.
-  basic_resolver_query(const basic_resolver_query& other)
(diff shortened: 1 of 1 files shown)


georgthegreat:
@chriskohlhoff could you, please, take a look?
We keep applying this patch locally and we would like to stop to.

georgthegreat:
@chriskohlhoff do you accept PR at all?

chriskohlhoff:
resolver::query was deprecated (asio 1.12.0, commit 74fe2b8e1) and removed (asio 1.33.0, commit e916bdfb1); the API this targets no longer exists. Closing.

georgthegreat:
The `basic_resolver_query` template still exists. Shouldn't it be removed too?

Issue #1493, opened 2024-06-17
ASIO provides no way to inject into pre-suspend/post-resume awaitable behavior or coroutine_specific_ptr


corporategoth:
I need to replicate something similar to thread-local storage in a project using coroutines.

Unfortunately, the awaitable class does not provide a way to do something prior to suspend in awaitable::await_suspend or post-resume in awaitable::await_resume()

Of course, the ideal would be able to have something like [boost::thread::thread_specific_ptr](https://www.boost.org/doc/libs/1_85_0/doc/html/thread/thread_local_storage.html#thread.thread_local_storage.thread_specific_ptr) that instead was more coroutine_specific_ptr instead.  Such that each awaitable_thread ends up with a different instance - so if I store something in the coroutine_specific_ptr, then what was in it before a co_await is also in it after co_await, but each coroutine stack could have different versions.

But lacking this, if I can inject into the await_suspend and await_resume I can implement a poor-man's version of this, by saving something from my thread_local somewhere in await_suspend, then restoring it on await_resume.
I started this by making my own version of awaitable, which then could contain the value I want to store.  Unfortunately, this breaks down very quickly, as co_spawn won't take my awaitable, and I also had to create my own version of awaitable_frame (inherited from the asio one) to add an await_transform for my awaitable, and also add coroutine_traits, etc.  ie. it just gets messier.

The ideal would be some kind of coroutine_specific_ptr that is part of asio, that could then more or less have it's backing store in the awaitable_thread, so that it could always just address the right one.

klemens-morgenstern:
Chris decided to make the `awaitable` a "walled of garden" on purpose, because it has strict requirements so that it can be used in multi-threaded environments.

[boost.cobalt](https://github.com/boostorg/cobalt/) (I am the author) only requires single-threading and everything else is pretty much open. I'd be happy to discuss ideas if that fits your needs.

corporategoth:
@klemens-morgenstern awaitable_specific_ptr works in a multithreaded environment.  And my environment IS multithreaded.  So there is no reason that awaitable_specific_ptr should not work (it works in my tests).

But even besides that - having access to some kind of coroutine_id (based on an ID from the lowest frame) would also be extremely helpful to be able to implement something outside of asio as well.

I am not about to switch a large commercial code base from a multi-threaded system to a single-threaded system, I would rather provide the enhancements to asio that it needs.

klemens-morgenstern:
[...]

PR #1496, opened 2024-06-18
Created asio::experimental::awaitable_specific_ptr

corporategoth:
Created asio::awaitable_specific_ptr


vipcxj:
I read your discussion in #1493. I think it would solve a lot of problems if we add a hook to the asio's awaitable context, such as issue in #1566

PR #1500, opened 2024-06-28
Fix `file_base::append` for both create/open for windows `stream_file`.

Ramirisu:
My previous PR didn't resolve the issue completely, so I create this to propose new fixes. The PR only affects `stream_file` for the windows platform.

Seek to the end of file when opening with flags:
1. w/r + append (open existing file) -> OPEN_EXISTING
2. w/r + create + append (open existing file or create if the file doesn't exist) -> OPEN_ALWAYS

Tested on Windows 10 22H2 19025.


chriskohlhoff:
Likely addressed by commit aa169b5e8, released in asio 1.32.0.

PR #1505, opened 2024-07-04
example: operations/composed_6: also free delay_timer_

ceggers-arri:
Also the delay_timer_ member should be freed before calling the user-supplied completion handler (as in the composed_7 and composed_8 examples).


chriskohlhoff:
Likely addressed by commit 595983d16, released in asio 1.31.0.

PR #1514, opened 2024-07-28
Error traits

klemens-morgenstern:
Other errors may be used by libraries (e.g. mysql or redis) to add more refined information. Supporting turning those into exceptions by default is however a requirement if library authors want a `void(my_error, ...)` completion signature.

### DIFF
--- a/asio/include/asio.hpp
+++ b/asio/include/asio.hpp
@@ -81,6 +81,7 @@
 #include "asio/dispatch.hpp"
 #include "asio/error.hpp"
 #include "asio/error_code.hpp"
+#include "asio/error_traits.hpp"
 #include "asio/execution.hpp"
 #include "asio/execution/allocator.hpp"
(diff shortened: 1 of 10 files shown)


anarthal:
This has been an absolute pain in Boost.MySQL, where getting error information provided by the server is really important for users. We created extra overloads for such information, but interoperability with Asio suffers. This PR would really address this need.

klemens-morgenstern:
Added as disposition in ecaa507b1b56484554ebc0cbcd300a483cf9a2e9.

Thanks Chris!

anarthal:
If the error type must be nothrow copy constructible, this is not a solution to neither Boost.MySQL nor Boost.Redis problems, as these use std::string for additional info.

zhscn:
Can we relax the restrictions and use is_nothrow_move_constructible?

PR #1518, opened 2024-08-19
Use swap idiom in move assign to correct destroy this

MBkkt:


MBkkt:
@chriskohlhoff is any chance you will looking on this? For now it's just leak when this move assign called

pfeatherstone:
Using `std::exchange()` would be better. You can set `other` to zero.

chriskohlhoff:
Likely addressed by commit ae93ce63b, released in asio 1.37.0.

PR #1529, opened 2024-09-20
Don't try to re-register regular file descriptors.

ceggers-arri:
File descriptors which have been rejected with EPERM by epoll_ctl() in register_descriptor() must also no be re-registered after fork.


chriskohlhoff:
Likely addressed by commit f53fa4a2f, released in asio 1.33.0.

Issue #1543, opened 2024-10-09
spawn() doesn't support move-only return values after boost 1.80.0


cbodley:
minimal reproducer fails to compile against the `boost-1.86.0` tag with both `gcc (GCC) 14.2.1` and `clang version 18.1.8` (both logs in https://gist.github.com/cbodley/1b3a85b6281133fc8ea96dfc1b9c2f5b)

```c++
#include <exception>
#include <iostream>
#include <memory>
#include <boost/asio.hpp>
#include <boost/asio/spawn.hpp>

int main()
{
  boost::asio::io_context ctx;
  boost::asio::spawn(ctx, [] (boost::asio::yield_context yield) {
        return std::make_unique<int>(42);
      }, [] (std::exception_ptr eptr, std::unique_ptr<int> result) {
        if (eptr) {
          std::rethrow_exception(eptr);
        } else {
          std::cout << *result << '\n';
        }
      });
  ctx.run();
  return 0;
}
```

see godbolt example in https://godbolt.org/z/bMjj11h7o

interestingly, this example does compile successfully against boost-1.80.0 where the feature was introduced:
> Changed spawn() to be a completion token-based asynchronous operation. This introduces new spawn() overloads that conform to the requirements for asynchronous operations.

but fails with any boost version after 1.80.0

the equivalent example with `co_spawn()` compiles and runs successfully:
```c++
#include <exception>
#include <iostream>
#include <memory>
#include <boost/asio.hpp>
#include <boost/asio/co_spawn.hpp>

int main()
{
  boost::asio::io_context ctx;
  boost::asio::co_spawn(ctx, [] () -> boost::asio::awaitable<std::unique_ptr<int>> {
        co_return std::make_unique<int>(42);                     
      }, [] (std::exception_ptr eptr, std::unique_ptr<int> result) {
        if (eptr) {
          std::rethrow_exception(eptr);
        } else {
          std::cout << *result << '\n';
        }
      });
  ctx.run();
  return 0;
}
```

PR #1546, opened 2024-10-18
test: add unit tests for spawn()

cbodley:
some simple test cases for spawn() with any_completion_handler, deferred, cancellation, and exceptions

the second test is a reproducer for https://github.com/chriskohlhoff/asio/issues/1543, and doesn't pass against master


cbodley:
> the second test is a reproducer for #1543, and doesn't pass against master

if you prefer, i'm happy to split this commit into a separate followup pr

PR #1556, opened 2024-11-05
co_composed.hpp: Guard use of async_operation

### LINKED ISSUE #1547: Asio from Boost 1.86.0 fails to build with C++20 w/o concepts
When porting my software to boost 1.86.0, I found a strange error that I  narrowed down to the following example.

`test.cpp`:

```c++
#include <boost/asio.hpp>

int main() { return 0; }
```

```
$ g++ -DBOOST_ASIO_DISABLE_CONCEPTS -std=c++20 test.cpp
In file included from /usr/include/boost/asio.hpp:69,
                 from test.cpp:2:
/usr/include/boost/asio/co_composed.hpp:849:13: error: ‘async_operation’ has not been declared
  849 |   template <async_operation Op>
      |             ^~~~~~~~~~~~~~~
/usr/include/boost/asio/co_composed.hpp:850:24: error: ‘Op’ has not been declared
  850 |   auto await_transform(Op&& op
      |                        ^~
[... and a few more errors, as compiler does not now what Op is]
```

I'm using x86_64 linux, gcc 13.2.1.

I see a similar error with clang 18.1.8.

If I drop  `-DBOOST_ASIO_DISABLE_CONCEPTS` or standard specification, the code compiles just fine.

iv-m:
`async_operation` is defined only when concepts are in use. When ASIO_DISABLE_CONCEPTS is defined, bare `typename` should be used instead.

We can't use ASIO_ASYNC_OPERATION macros here, as it currently does not have a no-argument form.

Fixes: #1547

### DIFF
--- a/asio/include/asio/co_composed.hpp
+++ b/asio/include/asio/co_composed.hpp
@@ -845,7 +845,11 @@ public:
     throw;
   }
 
+#ifdef ASIO_DISABLE_CONCEPTS
+  template <typename Op>
+#else
   template <async_operation Op>
+#endif
   auto await_transform(Op&& op
 #if defined(ASIO_ENABLE_HANDLER_TRACKING)
 # if defined(ASIO_HAS_SOURCE_LOCATION)


chriskohlhoff:
Likely already fixed / superseded by commit da2ad3cfe, released in asio 1.33.0.

Issue #1558, opened 2024-11-18
wrong offset passed to io_uring_prep_write_fixed/io_uring_prep_read_fixed for asio::stream_file


microcai:
in asio/detail/io_uring_descriptor_read_op.hpp and asio/detail/io_uring_descriptor_write_op.hpp

in do_prepare, it prepares the sqe by calling io_uring_prep_read_fixed/io_uring_prep_write_fixed

for  `offset` argument, it incorrectly pass 0 which should be -1.

microcai:
sent a pr for this issue https://github.com/chriskohlhoff/asio/pull/1559

chriskohlhoff:
Likely addressed by #1559, released in asio 1.34.0. Closing.

PR #1559, opened 2024-11-19
use -1 as correct offset for io_uring_prep_write_fixed/io_uring_prep_read_fixed

microcai:
this will address issue #1558


chriskohlhoff:
Likely addressed by commit 4c034a5f8, released in asio 1.34.0.

PR #1589, opened 2025-01-22
Fix promise_handler Allocator

justend29:
Although it's in the `detail` namespace, `experimental::detail::promise_handler` is very powerful for commanding promise completion from a callback chain.

Two small changes:

1. Correct the default `Allocator` template parameter.
2. Add a second constructor to `promise_handler` that provides a default allocator following the previously commented-out code.


chriskohlhoff:
Likely addressed by commit a9b5ac6e3, released in asio 1.34.1.

PR #1601, opened 2025-03-07
Annotate deprecated items with `[[deprecated]]`

### LINKED ISSUE #1571: Removal of deprecated things
Boost 1.87.0 (specifically ASIO 1.33[^1]) breaks https://github.com/zaphoyd/websocketpp by removing `io_service` and more (yes, I know about Boost.Beast).

It would be great if deprecated things were annotated with [`[[deprecated]]`](https://github.com/zaphoyd/websocketpp) so one can get compiler warnings _before_ the things are removed. If this gets too noisy, users could define `BOOST_ASIO_NO_DEPRECATION_WARNINGS` or something similar. Explicitly opting out is much better than requiring opt-in. I'm okay with removing stuff in minor releases[^2], but it's important to actively notify users about it - expecting users to use `BOOST_ASIO_NO_DEPRECATED` or to actively check the documentation for all their used functions is not enough in my opinion, we have `[[deprecated]]` for a reason.

[^1]: This release is not available in this repository, only in the Boost one.
[^2]: Ideally, the project would use semantic versioning, but I understand that this isn't possible.

Nerixyz:
Marks items disabled with `ASIO_NO_DEPRECATED` as `[[deprecated]]` (in C++ 14 and later). This helps users migrate earlier. Users can define `ASIO_IGNORE_DEPRECATED` to disable the warnings.

Fixes https://github.com/chriskohlhoff/asio/issues/1571.

### DIFF
--- a/asio/include/asio/basic_deadline_timer.hpp
+++ b/asio/include/asio/basic_deadline_timer.hpp
@@ -129,7 +129,7 @@ namespace asio {
 template <typename Time,
     typename TimeTraits = asio::time_traits<Time>,
     typename Executor = any_io_executor>
-class basic_deadline_timer
+class ASIO_DEPRECATED_X("Use basic_waitable_timer") basic_deadline_timer
 {
 private:
   class initiate_async_wait;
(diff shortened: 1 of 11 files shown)


chriskohlhoff:
Likely addressed by commit 9377ccbad, released in asio 1.37.0.

PR #1656, opened 2025-08-14
Fix memory leak in ssl::detail::engine move assignment

ashtum:


ashtum:
Fixed in https://github.com/chriskohlhoff/asio/commit/ae93ce63bb60114056a49271a52fcd9feb38bfdb.

PR #1658, opened 2025-08-14
add basic_yield_context::current

dnbazhenov:
basic_yield_context::current function allows to construct basic_yield_context object designated for the currently executed stackful coroutine.

This eliminates necessity to pass yield_context object from parent function to the invoked one. If the function is known to be always called in a context of a stackful coroutine, it can call basic_yield_context::current() passing either the IO context, or the context's  executor.

### DIFF
--- a/asio/include/asio/impl/spawn.hpp
+++ b/asio/include/asio/impl/spawn.hpp
@@ -62,6 +62,7 @@ public:
     : caller_(static_cast<fiber_type&&>(caller)),
       on_suspend_fn_(0),
       on_suspend_arg_(0)
+      caller_thread_(current_)
   {
   }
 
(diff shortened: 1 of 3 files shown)


dnbazhenov:
More description in
https://github.com/chriskohlhoff/asio/issues/1659

klemens-morgenstern:
Why don't you just do that with your own token, instead of adding overhead to every user?

dnbazhenov:
> Why don't you just do that with your own token, instead of adding overhead to every user?

single pair of load/store is non-significant. on the other hand, it's possible to put it under an option, if not needed.
passing yield_ctx over through every function as a parameter is greater overhead imo. better to construct it where it's actually needed

PR #1666, opened 2025-09-03
Fix #1423 resource leak in `awaitable::operator=`

### LINKED ISSUE #1423: Memory leak in asio::awaitable move assignment operator when lhs is valid
Example:

```cpp
// Clang 16, C++20, -fsanitize=address
#include <asio.hpp>

asio::awaitable<void> async_foo() {
  co_return;
}

asio::awaitable<void> async_main() {
  asio::awaitable<void> a{async_foo()};  // allocates here
  a = async_foo();  // leaks here
  co_await std::move(a);
}

int main() {
  asio::io_context io{1};
  asio::co_spawn(io, async_main(), asio::detached);
  io.run();
  return 0;
}
```

I didn't found any official document indicating such behavior, so I believe that is unintentional and needs a fix.

ddvamp:
`awaitable::operator=` did not destroy the existing frame

### DIFF
--- a/asio/include/asio/awaitable.hpp
+++ b/asio/include/asio/awaitable.hpp
@@ -80,7 +80,11 @@ public:
   awaitable& operator=(awaitable&& other) noexcept
   {
     if (this != &other)
+    {
+      if (frame_)
+        frame_->destroy();
       frame_ = std::exchange(other.frame_, nullptr);
+    }
     return *this;
   }
 


ddvamp:
Fixed in https://github.com/chriskohlhoff/asio/commit/0919c871f47fdc70628ac7c130f3f25ea141f68d

Issue #1328, opened 2023-07-17
Support high resolution waitable timers on Windows


mtnpke:
Windows traditionally uses a timer resolution of roughly 15.6 ms. This means that all timing-related functionality, such as waitable timers (that asio uses internally to implement all types of timers) or the `Sleep()` call, is limited to this resolution. If you call `Sleep(1)` or set a timer to expire after 1 ms and wait on it in a loop, the actual duration of the wait will be around 15.6 ms.

This might be sufficient for most programs, but there are use cases regarding, for example, real-time audio/video processing or communication, that can benefit vastly from a better timer resolution. Windows allows to improve the resolution by two ways:

* Using the (deprecated but still supported and widely used) Multimedia Timers API and, specifically, [timeBeginPeriod](https://learn.microsoft.com/en-us/windows/win32/api/timeapi/nf-timeapi-timebeginperiod) to request a better resolution of up to 1 ms for the whole process. This call may affect the OS and power consumption of the device as a whole and is generally a quite crude tool. Also, the resolution is limited to a static tick of 1 ms and multiples of it (so it does not help for waiting 1.5 ms). It can be called by the user independent of asio, though.
* Using the `CREATE_WAITABLE_TIMER_HIGH_RESOLUTION` flag for [CreateWaitableTimerEx](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createwaitabletimerexw). This is available since Windows 10 version 1803. It can be set for individual timers and therefore allows for much more fine-grained control. It is also not limited to a specific interval.

What do you think about supporting the HIGH_RESOLUTION flag in some fashion? I'm willing to put some effort into a PR if I get some direction on how this should behave from an API point of view. I'd certainly like to use that more in my application than having to resort to timeBeginPeriod().
Other projects such as Python are also beginning to support this API, see [this commit](https://github.com/python/cpython/commit/55868f1a335cd3853938082a5b25cfba66563135) for example.

NostraMagister:
The standard C++ library does not force an implementation to effectively provide nanosecond level timing and allows the high resolution clock to be mapped to, say, the steady clock. See NOTE in  https://en.cppreference.com/w/cpp/chrono/high_resolution_clock

[...]

PR #1685, opened 2025-11-23
High-res timers for Windows with IOCP support

### LINKED ISSUE #1404: Asio support for high resolution waitable timers Cross-platform
I would like to bring the subject of ticket  #1328 about Asio async high-resolution waitable timers support back under the attention but in a more cross platform context, since Asio is cross-platform.

Summarized the arguments for suggesting support for high resolution waitable timers in Asio are in essence that : 

- The standard C++ library does not force a C++ implementation to effectively provide nanosecond level timing and allows the high resolution clock to be mapped to, say, the steady clock. See NOTE in https://en.cppreference.com/w/cpp/chrono/high_resolution_clock

- Major Standard C++ implementation (e.g. MSVC Windows, GNU Linux) do not map the OS high resolution clock support but map the steady clock instead with its 15-25ms precision limitation, (15ms + a variable OS Thread Scheduler delay).

- High Res timer implementations are proprietary and poorly covered for cross platform : 
---> CreateWaitableTimerHighResolution() & CREATE_WAITABLE_TIMER_HIGH_RESOLUTION flag on Windows
---> timer_create() and timer_settime() functions with the CLOCK_MONOTONIC or CLOCK_REALTIME clock IDs for nanoseconds level timing on Linux and other OS that support the POSIX API. 

- Asio being an async IO library, it would be the library be excellence to 'even out' these shortcomings of the C++ specs & C++ implementations.
 
- As CPU/GPU/Bus IO/Networking speeds keep increasing, finer grained timeouts and timers become more important.

- Multi-media, gaming and 3D headset applications need better then 15ms precision timers for 60-90-120 fps.

- High speed lock-free atomic event queues would need 100 microsecond timeout loops in order not to loose their lock-free advantage.

For these reasons I would like to attract the attention of this to the makers of this great Asio library and see if someone could take a positions as to this subject (such as - we will never add this, we could add this, we will add this,...). This would allow developers to know if they need to look for other solutions, possibly temporary, or if they can keep using the Asio async. high_resolution_clock C++ Standard notation timers, knowing that Asio will map the object to a real high resolution clock.

I would also like to call for support from other posters to find out what their take is on the high resolution timers subject as i suspect that many developers would be served by using Asio with cross platform async. timers that are really based on an high resolution clock instead of on a mapped lower resolution clock.

Thank you.

reinar:
This PR adds high-resolution timer support for Windows ([doc](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createwaitabletimerexw), available from Windows 10, version 1803) and if such support is enabled - associates it with iocp port to get rid of separate timer thread.

Feature is gated behind `ASIO_ENABLE_IOCP_HIRES_TIMERS` define, so default behavior won't be changed from previous versions of asio.

Both on and off states for the feature were tested on Windows 11 26100.7171 with MSVC 19.50.35718 and clang-cl 21.1.6 (default unit test setup, just overridden with boost 1.89.0. Unit tests are passing).

Fixes #1404, #1328

### DIFF
--- a/include/asio/detail/config.hpp
+++ b/include/asio/detail/config.hpp
@@ -809,6 +809,17 @@
 # endif // defined(ASIO_WINDOWS) || defined(__CYGWIN__)
 #endif // !defined(ASIO_HAS_IOCP)
 
+#if !defined(ASIO_HAS_IOCP_HIRES_TIMERS)
+# if defined(ASIO_HAS_IOCP)
+#  include <sdkddkver.h>
+#  if defined(NTDDI_VERSION) && (NTDDI_VERSION >= NTDDI_WIN10_RS4)
+#   if defined(ASIO_ENABLE_IOCP_HIRES_TIMERS)
+#    define ASIO_HAS_IOCP_HIRES_TIMERS 1
+#   endif // defined(ASIO_ENABLE_IOCP_HIRES_TIMERS)
+#  endif // defined(NTDDI_VERSION) && (NTDDI_VERSION >= NTDDI_WIN10_RS4)
+# endif // defined(ASIO_HAS_IOCP)
+#endif // !defined(ASIO_HAS_IOCP_HIRES_TIMERS)
+
 // On POSIX (and POSIX-like) platforms we need to include unistd.h in order to
 // get access to the various platform feature macros, e.g. to be able to test
 // for threads support.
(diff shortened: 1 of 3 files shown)


mtnpke:
@reinar ping - what do you think about setting period = 0?

Issue #1690, opened 2025-12-03
Missing outbound_interface creating via address_v6


marek22k:
The code
```
    const boost::asio::ip::address_v4 addr = endpoint.address().to_v4();
    const boost::asio::ip::multicast::outbound_interface oif(addr);
    this->_socket.set_option(oif);
```
compiles, but not
```
    const boost::asio::ip::address_v6 addr = endpoint.address().to_v6();
    const boost::asio::ip::multicast::outbound_interface oif(addr);
    this->_socket.set_option(oif);
```
It would be nice to have full IPv6-Support.

pinwhell:
@chriskohlhoff if asio would like to support `address_v6 -> outbound_interface` would you consider this a bug, missing feature or "by design"? how would a solution for this looks like for you? maybe introducing logic dependent on platform or API change? 🤔

PR #1702, opened 2026-01-16
Clarify IPv4 vs IPv6 semantics of multicast::outbound_interface

pinwhell:
`multicast::outbound_interface` conceptually represents a network interface.

In IPv4, allowing construction from an `address_v4` works because an address often uniquely identifies an interface.
This assumption does not hold for IPv6, where an address does not uniquely identify a specific interface.

This change does not alter behavior, but clarifies the intended semantics by documenting that IPv6 requires specifying the interface index rather than an IPv6 address.

Related #1690.


chriskohlhoff:
Likely addressed by commit 8a22a0ecb, released in asio 1.38.1.

PR #1710, opened 2026-01-29
Change posix_thread entry symbol name to avoid clashes #1709

### LINKED ISSUE #1709: Allocator-aware implementation in posix_thread leads to crashes in dynamically loaded software
We experience crashes when loading plugin code (i.e. a .dylib on macOS) into host processes that have been built against an earlier version of boost::asio. 
Before [ca3cabafde6fb3daf12255d721f2a8374a1991ec](https://github.com/chriskohlhoff/asio/commit/ca3cabafde6fb3daf12255d721f2a8374a1991ec) the implementation of posix_thread was deleting its argument before return. In the current implementation this task is performed by the thread that calls `::pthread_join` .

In an environment where a host process that is compiled with an asio version before above commit loads a plugin (i.e. .dylib) that is compiled against the new version the resolution of `asio_detail_posix_thread_function` is ambiguous and ultimately leads to a double-free of its argument.

A simple fix is to rename the thread-entry C-function (e.g. `asio_detail_posix_thread_function_v2` ). 

Reproducible (at least) on macOS 26.2 with Xcode 26.2 in at least 2 commercially available host-software products.

EikeAtOT:
Add suffix to posix thread entry function to avoid name clashes with older library versions.
Fixes #1709

### DIFF
--- a/boostify.pl
+++ b/boostify.pl
@@ -131,9 +131,9 @@ sub copy_source_file
       $line =~ s/[\\@]ref async_read/boost::asio::async_read()/g;
       $line =~ s/[\\@]ref async_write/boost::asio::async_write()/g;
     }
-    if ($line =~ /asio_detail_posix_thread_function/)
+    if ($line =~ /asio_detail_posix_thread_function_for_external_allocator/)
     {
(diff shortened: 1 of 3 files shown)


chriskohlhoff:
Likely addressed by commit 8f1ecef92, released in asio 1.38.1.

PR #1713, opened 2026-02-11
Annotate intended fall-through cases

orgads:


chriskohlhoff:
Likely addressed by commit 6a6e02ce0, released in asio 1.38.1.

PR #1717, opened 2026-02-26
Fix MSG_NOSIGNAL handling on NetBSD.

0-wiz-0:
In /usr/include/sys/unistd.h:62, `_POSIX_VERSION' is only `200112L', but NetBSD supports MSG_NOSIGNAL anyway.


chriskohlhoff:
Likely addressed by commit 818477578, released in asio 1.38.1.

PR #1730, opened 2026-03-29
Fix clang-cl build

hukeyue:


chriskohlhoff:
Likely addressed by commit ed89dac75, released in asio 1.38.2.

PR #1739, opened 2026-04-16
Add ASIO_INLINE_VARIABLE to variables in verify_mode.hpp

Joker-Principal:
Allow user to export these symbols in modules, see also https://discourse.llvm.org/t/modules-how-to-export-variables-in-the-gmf/72083


Joker-Principal:
@chriskohlhoff Although I’ve noticed that very few pull requests get accepted, I believe this one is simple enough to review in two minutes.

PR #1750, opened 2026-05-29
Feat/modsup

kydos:
I've had some challenges in using ASIO with a C++23 project that heavily uses modules when compiling with GCC.  These PR addresses the problems I've seen. Happy to hear your comments.


chriskohlhoff:
Likely addressed by commit efc2f7754, released in asio 1.38.2.

kydos:
Thanks, will give it a try!

PR #1753, opened 2026-06-07
[[maybe_unused]] annotation vs. commented-out parameters

toni1254:
When going through the tutorials section I realized that unused input parameters are usually commented out. Since (I think) C++17 the `unused_maybe` annotation could be used though. Thus, I was wondering whether there is any specific reason for commenting out params in asio rather than using such "modern" semantics.

### DIFF
--- a/src/doc/tutorial.qbk
+++ b/src/doc/tutorial.qbk
@@ -165,7 +165,7 @@ Using asio's asynchronous functionality means supplying a [link asio.overview.mo
 
 
 
-  ``''''''``void print(const std::error_code& /*e*/)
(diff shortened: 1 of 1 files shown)


annihilatorq:
As you said, `[[maybe_unused]]` is a C++17 attribute, while [asio's minimum supported standard](https://think-async.com/Asio/AsioStandalone.html) is C++11. When compiling in pre-C++17 mode this will probably produce unknown attribute warnings, and some compilers may not be able to even parse `[[...]]` at all, so the tutorial code would no longer compile on the oldest supported toolchains.

###### This is just my view, so please don't take my comment as some kind of PR review

PR #1754, opened 2026-06-11
asio: io_uring: add ASIO_DISABLE_IO_URING_IOWAIT

jaehyoo:
The io_uring backend marks the io_context waiter task as in_iowait whenever any SQE is pending. The kernel doesn't distinguish real I/O SQEs from POLL_ADD/TIMEOUT SQEs submitted for socket and timer waits, and the backend keeps those pending for the full async_wait, so the flag stays set with nothing blocked on I/O. This inflates iowait/load metrics and blocks deeper CPU idle states; on platforms without cpufreq scaling the original hint is moot but the accounting damage remains.
  
Add an opt-in macro ASIO_DISABLE_IO_URING_IOWAIT calling io_uring_set_iowait(ring, false) after queue init; default behaviour is unchanged. The call and the IORING_FEAT_NO_IOWAIT flag both landed in liburing 2.12, so it is guarded on the flag (#error otherwise). Older runtime kernels return -EOPNOTSUPP, which is harmless.

### DIFF
--- a/include/asio/detail/impl/io_uring_service.ipp
+++ b/include/asio/detail/impl/io_uring_service.ipp
@@ -538,6 +538,14 @@ void io_uring_service::init_ring()
     asio::detail::throw_error(ec, "io_uring_queue_init");
   }
 
+#if defined(ASIO_DISABLE_IO_URING_IOWAIT)
+# if !defined(IORING_FEAT_NO_IOWAIT)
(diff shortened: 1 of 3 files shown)


chriskohlhoff:
Likely addressed by commit ea7988e7c, released in asio 1.38.2.

PR #1761, opened 2026-08-10
fix tsan findings and compiler warnings

### LINKED ISSUE #1193: GCC+TSan error
Compiling Asio with GCC+TSan gives me the following error:

```
In function 'void std::atomic_thread_fence(memory_order)',
    inlined from 'boost::asio::detail::std_fenced_block::std_fenced_block(full_t)' at ./boost/asio/detail/std_fenced_block.hpp:46:29:
/usr/include/c++/12/bits/atomic_base.h:133:26: error: 'atomic_thread_fence' is not supported with '-fsanitize=thread' [-Werror=tsan]
  133 |   { __atomic_thread_fence(int(__m)); }
      |     ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
In function 'void std::atomic_thread_fence(memory_order)',
    inlined from 'boost::asio::detail::std_fenced_block::~std_fenced_block()' at ./boost/asio/detail/std_fenced_block.hpp:52:29:
/usr/include/c++/12/bits/atomic_base.h:133:26: error: 'atomic_thread_fence' is not supported with '-fsanitize=thread' [-Werror=tsan]
  133 |   { __atomic_thread_fence(int(__m)); }
      |     ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
cc1plus: all warnings being treated as errors
```

Because TSan does not support `atomic_thread_fence`.

LLVM's https://github.com/llvm/llvm-project/blob/ad1161f9b5ff251b80788033e4db82b5a11b187b/compiler-rt/include/sanitizer/tsan_interface.h#L24-L25 can be used instead. 

I'm not sure there's anything actionable about this but I'm just reporting it.

I've also noticed `BOOST_ASIO_DISABLE_FENCED_BLOCK` is there. I imagine disabling the fenced block takes care of the GCC warning, but if these fences are really necessary, TSan will report data races.

d3matt:
use tsan intrinics rather than std::atomic_thread_fence when built with -fsanitize=thread with the latest versions of gcc, fixes #1193


jeremy-murphy:
yes! yes! yes! yes! yes!

@chriskohlhoff, could you **please** give this consideration? It's hugely impractical to use Thread Sanitizer without it.

moha-gh:
Would also love to see this merged!

PR #1768, opened 2026-09-15
Check descriptor validity before kqueue registration

onalante-ebay:
Otherwise, no operations on the descriptor will complete since the
return code of `::kevent` for non-speculative, non-read-only operations
is unchecked[^1].  `register_descriptor_internal` does not need guarding
since it is only passed library-instantiated descriptors.

For comparison, the epoll and IOCP backends immediately fail
`register_descriptor` and `register_handle` (respectively) when given an
invalid descriptor, while the io_uring backend handles errors uniformly
through CQEs.

[^1]: https://github.com/chriskohlhoff/asio/blob/8806a6803cde7054c3049d3666d3ec36786568c5/include/asio/detail/impl/kqueue_reactor.ipp#L274


onalante-ebay:
Opting to add a check to the `::kevent` call in `start_op` instead to prevent having a syscall in the happy path.

PR #1769, opened 2026-09-15
Check `::kevent` result when non-{speculative,read-only}

onalante-ebay:
Otherwise, no operations on the descriptor will complete since the
return code of `::kevent` for non-speculative, non-read-only operations
is unchecked[^1].

[^1]: https://github.com/chriskohlhoff/asio/blob/8806a6803cde7054c3049d3666d3ec36786568c5/include/asio/detail/impl/kqueue_reactor.ipp#L274


onalante-ebay:
Supersedes #1768.

PR #1773, opened 2026-09-21
Replace from_string with make_address in documentation examples.

dornbirndevelops:
The `ip::address::from_string` and `ip::address_v4::from_string` functions were deprecated in favour of `make_address` and `make_address_v4`, and removed in asio 1.34.0.
Several code snippets still used them, so the examples no longer compile as written.
Update those snippets to the current free functions. No library code is affected.


dornbirndevelops:
Should the usages in `reference.qbk` also be adjusted?

